Issue #1, opened 2022-04-25
[RFC] General Attestation Service Design Proposal


liangzhou121:
# Design
According to [[RFC] Generic Key Broker Service (KBS) & Attestation Service high level architecture proposal](https://github.com/confidential-containers/documentation/issues/30), CC's general Attestation system relies on Attestation Service and KBS to verify the Attestation-Agent's running environment and distribute critical messages(such as KEK). The KBS is the role of RATS `Relying Party`, it's hardware agnostically and focuses on vendor-specific requirements. The Attestation Service is an implementation of the RATS `Verifier` role, it focuses on the verification of Evidence's identity and TCB status.

And the CC's Attestation Service is a general infrastructure service that mostly focuses on providing a Evidence verification service to KBS. So it needs to be compatible with all CC-supported HW-TEEs and should have excellent scalability to support potential new HW-TEE in the future. It also can be compatible with existing third-party attestation services with modularization.

To support the verification of TCB status, Attestation Service has the capability of parse the received Evidence to extract its corresponding TCB status. At the same time, it also includes the Policy Engine component which is an implementation of RATS `Verifier Owner`, It relies on the reference data from `Reference Value Provider` and the info extracted from Evidence to match the Attestation-Agent's running environment's TCB status dynamically. And generate the corresponding Attestation Results.

Attestation Service supports different deployment modes, as a service or as a library. One Attestation Service maybe needs to support several KBS services simultaneously when it is deployed as an infrastructure service. This will enhance the flexibility of the service's deployment.

## Goals
- Define the format of Evidence and Attestation Results messages for SGX and TDX.
- Distribute the Attestation Service as a library or single service.
- It's a general infrastructure service that:
	- It should compatible with various HW-TEEs and have good scalability.
	- It should compatible with different third-party attestation Services.
- Support management API to set reference data by `Reference Value Provider` and set other configurations.

## Non-Goals
- Define the format of Evidence and Attestation Results messages for SEV(-ES) and SEV-SNP.
- Proxy sub-modules that support existing third-party attestation services.

# Architecture
The following diagram demonstrates the overall architecture and internal modules:
![image](https://user-images.githubusercontent.com/83809742/165093645-b554ad79-f988-4c8f-915e-50c49dfc4cae.png)
- Service: Provides interface that can be used to communicate with KBS. It receives Evidence from KBS and returns the Attestation Results.
- Attestation: The core module that implements Evidence's attestation.
[...]

PR #2, opened 2022-05-12
AS: The initial implementation

liangzhou121:
1.A Crate that implements the Attestation Service core functionality.

Fixes: #1
Signed-off-by: zhouliang121 <liang.a.zhou@linux.alibaba.com>


lib/src/core/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+mod proxy;
+pub mod verifier;
+
+#[async_trait]
+pub trait Verifier {
+    async fn evaluate(&self, evidence: &Value) -> Result<String>;
+}
+
+#[derive(Debug)]
+pub struct Core {
```
fitzthum, line 16: `Core` is a rather vague name. Could we just call it `AttestationService` or something else a little bit more descriptive?

lib/src/core/verifier/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+pub mod policy;
+mod sample;
+mod sgx;
+mod tdx;
+
+#[derive(Debug)]
+pub struct Evidence {
+    pub quote: String,
+    pub ehd: String,
+    pub add: String,
```
fitzthum, line 16: I think the RFC Issue has `aad` for this field rather than `add`. I assume this refers to `Additional Authenticated Data`?

lib/src/core/verifier/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+}
+
+#[derive(Debug)]
+pub struct AttestationResults {
+    pub tee: String,
+    pub attestation_type: String,
```
fitzthum, line 22: Why do we need both a `tee` and `attestation_type` field? Seems like these would be the same value?

lib/src/core/verifier/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+            "sample" => Ok(self.sample.evaluate(&self.sample.parse(evidence)?).await?),
+            "sgx" => Ok(self.sgx.evaluate(&self.sgx.parse(evidence)?).await?),
+            "tdx" => Ok(self.tdx.evaluate(&self.tdx.parse(evidence)?).await?),
+            _ => Err(anyhow!("not supported!")),
+        }?;
+        attestation_results(&results)
+    }
+}
+
+fn attestation_results(results: &AttestationResults) -> Result<String> {
```
fitzthum, line 68: It might be more idiomatic to make an `AttestationResults` struct that is returned by the evaluate method of each verifier. Then you could just serialize the struct here.

lib/src/core/verifier/sample/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+                .ok_or(anyhow!("Extract <is_debuggable> failed"))?,
+            cpusvn: v["cpusvn"]
+                .as_u64()
+                .ok_or(anyhow!("Extract <cpusvn> failed"))?,
+            svn: v["svn"].as_u64().ok_or(anyhow!("Extract <svn> failed"))?,
+        };
+        let ev = Evidence {
+            quote: serde_json::to_string(&quote)?,
+            ehd: "".to_string(),
+            add: "".to_string(),
```
fitzthum, line 59: Why do this parsing manually? Can't you just deserialize the JSON directly into the evidence struct?

lib/src/core/verifier/sample/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+
+        let tcb = tcb_status(&evidence.quote)?;
+        let input = json!({
+            "cpusvn": tcb.cpusvn,
+            "svn": tcb.svn
+        })
+        .to_string();
+        let policy = get(self.opa_policy_path()?.as_path()).await?;
+        let reference = get(self.opa_reference_path()?.as_path()).await?;
+        let evaluation = evaluate(policy, reference, input)?;
```
fitzthum, line 77: Where does this `evaluate` function come from?

lib/src/core/verifier/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+    pub policy: String,
+    pub add: String,
+}
+
+#[async_trait]
+pub trait TEEtraits {
+    fn parse(&self, evidence: &Value) -> Result<Evidence>;
+    async fn evaluate(&self, evidence: &Evidence) -> Result<AttestationResults>;
+    fn opa_policy_path(&self) -> Result<PathBuf>;
+    fn opa_reference_path(&self) -> Result<PathBuf>;
```
fitzthum, line 32: Might be clearer to call this `opa_reference_data_path`

fitzthum (commented):
Looks good so far. I will add more comments soon. 

There are a number of places where I think the JSON parsing could be simplified.

lib/build.rs
```diff
@@ -0,0 +1,34 @@
...
+    if !opa.status.success() {
+        return Err(std::str::from_utf8(&opa.stderr.to_vec())
+            .unwrap()
+            .to_string());
+    }
+
+    Ok(())
+}
+
+fn main() {
```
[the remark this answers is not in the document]
liangzhou121: This wrapper is used to interrupt and output the `build.rs` execution's error message if any error happens. I also refer the implemenation of `kata-containers\src\libs\protocols\build.rs` file's `main()` wrapper.

lib/src/core/verifier/sample/mod.rs
liangzhou121: Yes, use the `#[derive(Serialize, Deserialize, Debug)]` traits are better.

lib/src/core/verifier/sample/mod.rs
liangzhou121: This `evalute()` is from OPA module. I will switch it to `opa::evalute()` explicitly.

lib/src/core/verifier/mod.rs
```diff
@@ -0,0 +1,46 @@
...
+            tdx: tdx::Tdx::new(workdir.clone()).await?,
+            sgx: sgx::Sgx::new(workdir.clone()).await?,
+            sample: sample::Sample::new(workdir.clone()).await?,
+        })
+    }
+}
+
+#[async_trait]
+impl Verifier for Verifiers {
+    async fn evaluate(&self, evidence: &Evidence) -> Result<AttestationResults> {
```
fitzthum, line 38: It seems like this functionality could just go in the `Attestation` struct. Do we really need to have a separate `Verifiers` struct?

lib/src/core/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+use async_trait::async_trait;
+
+mod proxy;
+pub mod verifier;
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Evidence {
+    pub tee: String,
+    pub quote: String,
+    pub ehd: String,
```
fitzthum, line 14: The KBS needs access the nonce field of the evidence to make sure that the correct nonce was returned from the TEE. In the RFC the nonce is part of the `ehd` but here we don't explicitly specify the format of the `ehd`. We might want to create an `ehd` struct that defines the `ehd` field of the evidence struct so that we guarantee the nonce is always in the same place. Another option would be to add a `nonce` method to `TEETraits` that the KBS could call to get the nonce.

lib/src/core/verifier/sgx/policy.rs
```diff
@@ -0,0 +1,69 @@
...
+        let path = self.opa_path()?;
+        if !path.exists() {
+            fs::create_dir_all(path).await?;
+        }
+
+        let file = self.opa_policy_path()?;
+        if !file.exists() {
+            info!("{} isn't exist", file.to_str().unwrap());
+            let policy = r#"
+package policy
```
fitzthum, line 23: Why not just include some default policies in the repo? Do we really need these functions to add the policies? It seems like this is something we only need to do once when we first install the AS.

fitzthum (commented):
A few more comments. It might be a good idea to split the `lib` and `server` stuff into two PRs.

lib/src/core/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+    pub aad: String,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct AttestationResults {
+    pub tee: String,
+    pub policy: String,
+    pub aad: String,
```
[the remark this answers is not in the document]
liangzhou121: According to it's proposal's description, this field actually includes the TEE tcb status. So I will rename the `add` -> `tcb`.

lib/src/core/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+}
+
+#[async_trait]
+pub trait Verifier {
+    async fn evaluate(&self, evidence: &Evidence) -> Result<AttestationResults>;
```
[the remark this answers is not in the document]
liangzhou121: Yes, this `Verifier` trait defines these attestation related interfaces that are exported by the verifier module. So I will move it to the `lib/src/core/verifier/mod.rs`

lib/src/core/verifier/mod.rs
liangzhou121: I will add a new trait named `opa_files_path()` to export these specific TEE's OPA related interfaces unified. And I think better to keep this trait, because maybe other new trait functions will be added in future.

lib/src/core/mod.rs
liangzhou121: OK. I will extend the `sample` TEE codes to add the `ehd` definition example.
We can ensure the SGX, TDX, SEV-SNP attestation needs the nonce, but can't sure whether any new TEE will also depends on the nonce, so I think it's inapporiate to add the `ehd` explicit definition in this `core/mod.rs`.

lib/src/core/verifier/sgx/policy.rs
liangzhou121: The background is this `AttesattionService` Crate has the features:
- It can be distributed as a single Crate.
- It should also support the potential multiple users scenario, its initial function's input includes a `user_id`, so different users will have their own policy.rego and reference files. And these files' path should also be different.
```
pub async fn new(path: &Path, user_id: String) -> Result<Self> {}
```

> Why not just include some default policies in the repo?

I think it's difficult to deploy these default policy and reference files via command directly (such as `make install`) in this multiple users scenario.

> Do we really need these functions to add the policies? It seems like this is something we only need to do once when we first install the AS.

I think they are needed, they will be invoked each time when any new `user_id` coresponding `AttesattionService` instance is created.

lib/src/core/mod.rs
fitzthum: If some future architecture comes along, it must support a nonce or it won't work with the KBS. If there is no nonce, there is no way for the KBS to match the challenge to the evidence. We need to be flexible about the format of the evidence, but I think it is fine and maybe even good if we require all future evidence to include a nonce.

lib/src/core/verifier/sgx/policy.rs
fitzthum: Ok, I see where you're coming from.

If these are just default policies and reference data, couldn't they be shared between all users? We could have one directory that stores all the defaults (which are tracked in git as `.rego` file) and other directories could be created for each user.

I also wonder if the lib is the right place to handle manipulating files like this. It seems like maybe the app should be in charge of maintaining all per-user policy state. By the time the verifier is called all the policies should be in the right places.

lib/src/core/verifier/sample/policy.rs
```diff
@@ -0,0 +1,49 @@
+use super::*;
+use anyhow::Result;
+use std::path::PathBuf;
+use tokio::fs;
+
+impl Sample {
+    pub fn opa_path(&self) -> Result<PathBuf> {
```
fitzthum, line 7: Do you really need this method? Seems like you could just calculate the path directly below.

lib/src/core/verifier/sample/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Quote {
+    is_debuggable: bool,
+    cpusvn: u64,
+    svn: u64,
+    report_data: String,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Policy {
```
fitzthum, line 36: We're defining the same `Policy` struct for every verifier? It seems like this could be moved to the core file. The name policy is also a little bit confusing. This struct doesn't represent a policy, but the result of evaluating a policy.

lib/src/core/verifier/sample/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+    allow: bool,
```
fitzthum, line 37: Can we change this to a string in order to support more than just allow/reject?

lib/src/core/verifier/sample/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+    name: String,
+    hash: String,
+    diagnose: String,
```
fitzthum, line 40: `name` -> `policy_name` It seems like all the policies have the same name anyway? 
`hash` -> `policy_hash` Not clear what this is the hash of otherwise.
`diagnose` -> `info`

lib/src/core/verifier/sample/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+    }
+    Ok(())
+}
+
+fn attestation_results(tcb: &Tcb, evaluation: String) -> Result<AttestationResults> {
+    let v: Value = serde_json::from_str(&evaluation)?;
+
+    let policy = Policy {
+        allow: v["allow"].as_bool().ok_or(anyhow!("convert allow error"))?,
+        name: "policy.rego".to_string(),
```
fitzthum, line 115: So we only support one policy file at a time and it must be called policy.rego?

fitzthum (commented):
A few more comments.

lib/src/core/verifier/sgx/policy.rs
liangzhou121: Yes, I am think you are right, it's reasonable to leave these files' creation to app program (such as `server`) totally. After the `AttestationService` instance is created by app, the app needs to inject these dependent policy files explicitly (Like the following codes).
```
let service = AttestationService::new(&Path::new("./"), user_id.clone()).await?;
service.policy_engine_set_file(<set policy file cmd>).await?;
service.policy_engine_set_file(<set reference data file cmd>).await?;
```

It can also help to keep the lib's logic as simple/clear as possible.

lib/src/core/verifier/sample/mod.rs
liangzhou121: No, this struct is TEE specific, so it can be extended to change and add any other info. If the some TEE's evaluation is executed by multiple policy files, it maybe like this:
```
#[derive(Serialize, Deserialize, Debug)]
pub struct Evaluation {
    allow: String,
    policy_name1: String,
    policy_hash1: String,
    policy_name2: String,
    policy_hash2: String,
    info: String,
}
```

lib/src/core/mod.rs
liangzhou121: I think i can catch your point, so I tried to update the Evidence struct as following:
```
// core/mod.rs

#[derive(Serialize, Deserialize, Debug)]
pub struct Ehd {
    // The mandate nonce
    pub nonce: String,
    // The TEE specific enclave hold data
    pub ehd: String,
}
#[derive(Serialize, Deserialize, Debug)]
pub struct Evidence {
    pub tee: String,
    pub quote: String,
    pub aad: String,
    pub ehd: Ehd,
}
```
But this will cause the error: _Deserialize Evidence error, expected struct Ehd at line 1 column 116)_ during the input evidence's deserialize.
This means we need to parse the json evidence manully to construct the `Evidence` struct. So I think we need to keep the `Ehd` struct's definition in each specific verifier.

lib/src/core/verifier/sgx/policy.rs
liangzhou121: Hi @fitzthum ,
I think this problem again and have the following new answers:

> If these are just default policies and reference data, couldn't they be shared between all users? We could have one directory that stores all the defaults (which are tracked in git as .rego file) and other directories could be created for each user.

Yes, these policy and reference files are shared by all users. And I will update the codes to make these files are tracked in git as `.rego`/`.json` files. 
And I think it should be better to put these files to their corresponding TEE folder. It makes the structure clear.

> I also wonder if the lib is the right place to handle manipulating files like this. It seems like maybe the app should be in charge of maintaining all per-user policy state. By the time the verifier is called all the policies should be in the right places.

I think it should be the AS Crate's responsibility to provide these default `policy.rego` and `reference_data.json` files. It's too hard for the Crate's new user to provide the `policy.rego` by themselves to make it work correctly. 
And the AS's default `policy.rego` is also a reference policy if the user want to update and enhance the policy later.

I also hunch that most AS Crate users won't care the `policy.rego` file, they will update the `reference_data.json` only.

lib/src/core/mod.rs
fitzthum: I was thinking about this more and we might actually need to do something similar for the TEE public key. The KBS needs to access to the key in order to encrypt the secrets, but we also have to make sure that the key that the KBS thinks is the public key of the TEE matches the public key in the evidence. Every platform must have some way to link the TEE public key to the attestation report (just like the nonce).

Interestingly the nonce and public key are both provided in two places. 

First, the KBS protocol specifies the attestation as 
```
{
    "evidence": {
	    "nonce": "",
	    "tee-pubkey": {
		    "algorithm": "",
		    "pubkey-length": "",
		    "pubkey": ""
		},
	    "tee-evidence": {}
	}
}
```
Then, the AS gives this example of `evidence`
```
{
    "tee" : "tdx",
    "quote": <Base64 encoded quote>,
    // The ehd's hash will be embeded in Quote->report_data.
    "ehd": 
    {
        "nonce" : "cdfunsas23445xd",
        "public-key" : "xxxxxxxx"
    },
    "aad": 
    {
        "tdelInfo" : <Base64 encoded eventlog's information>,
        "tdelData" : <Base64 encoded eventlog's data>
    }
}
```

In order for this to be secure, I think that the KBS or the AS has to make sure that the nonce and public-key in the evidence exactly match the nonce and the public key in the attestation. There are two options. First, the KBS could check that the values match. For this to work we need to make sure that the evidence always includes the nonce and the public key in the same place. This is what I first suggested. There is another options that might be nicer. We could have the AS check that the values match instead of the KBS. All we would need to do is pass the entire attestation payload into the AS rather than just the evidence. if we pass the entire attestation payload (which btw should probably include an `additional-params` field) into the AS, we don't have to mandate the format of the evidence at all. We just have to make sure that the AS checks that the nonce and public key match the (signed) nonce and public key in the attestation report. 

Also note that in the example above the evidence does not contain the same key metadata that the attestation does. Is this an issue?

lib/src/core/verifier/sample/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+struct Quote {
+    is_debuggable: bool,
+    cpusvn: u64,
+    svn: u64,
+    report_data: String,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+struct Evaluation {
+    allow: String,
```
fitzthum, line 38: also, `allow` -> `result`

lib/src/core/verifier/policy/opa/mod.rs
```diff
@@ -0,0 +1,109 @@
+use anyhow::{anyhow, Result};
+use std::ffi::CStr;
+use std::os::raw::c_char;
+
+/// Link import cgo function
+#[link(name = "opa")]
+extern "C" {
+    pub fn evaluateGo(policy: GoString, data: GoString, input: GoString) -> *mut c_char;
+}
+
```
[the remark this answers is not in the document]
liangzhou121: Yes, I couldn't find a mature OPA Crate currently. The only one I can google is the following that in some guy's private [repo](https://github.com/myagley/opa-rs). Maybe we can switch to a mature OPA Crate in future.

lib/src/core/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+    pub tee: String,
+    pub quote: String,
+    pub ehd: String,
+    pub aad: String,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct AttestationResults {
+    pub tee: String,
+    pub evaluation: String,
```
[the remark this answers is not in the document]
liangzhou121: This is a JSOn string and should include the OPA's evaluation results and the policy's info which is used by OPA.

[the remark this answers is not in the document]
sameo: I think it's safe to say that `AttestationResults` will always carry the policy and the policy's evaluation results, but that they may have different formats depending on the TEE? If that's so, could we split that field into a `policy` and a `result` fields?

lib/src/core/verifier/sgx/policy.rs
```diff
@@ -0,0 +1,29 @@
...
+use anyhow::Result;
+use tokio::fs;
+
+impl Sgx {
+    // Set the opa policy engine's working directory.
+    pub async fn default(&self) -> Result<()> {
+        let path = self.workdir.join("opa").join("sgx");
+        if !path.exists() {
+            fs::create_dir_all(path).await?;
+        }
```
dubek, line 11: you may remove the surrounding `if`.

lib/src/core/verifier/tdx/policy.rs
```diff
@@ -0,0 +1,29 @@
...
+use anyhow::Result;
+use tokio::fs;
+
+impl Tdx {
+    // Set the opa policy engine's working directory.
+    pub async fn default(&self) -> Result<()> {
+        let path = self.workdir.join("opa").join("tdx");
+        if !path.exists() {
+            fs::create_dir_all(path).await?;
+        }
```
dubek, line 11: you may remove the surrounding `if` condition.

lib/src/core/verifier/sample/policy.rs
```diff
@@ -0,0 +1,29 @@
...
+use anyhow::Result;
+use tokio::fs;
+
+impl Sample {
+    // Set the opa policy engine's working directory.
+    pub async fn default(&self) -> Result<()> {
+        let path = self.workdir.join("opa").join("sample");
+        if !path.exists() {
+            fs::create_dir_all(path).await?;
+        }
```
dubek, line 11: the `if` is not needed.

lib/src/core/verifier/policy/opa/opa.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/open-policy-agent/opa/storage/inmem"
+)
+
+//export evaluateGo
+func evaluateGo(policy string, data string, input string) *C.char {
+	// Deserialize the message in json format
+	input_map := make(map[string]interface{})
+	err := json.Unmarshal([]byte(input), &input_map)
+	if err != nil {
+		return C.CString("Error:: " + err.Error())
```
dubek, line 20: this function returns two types of CStrings: if it begins with `Error:: ` then it's an error, otherwise it's a "good" result. Instead, I suggest the following:

1. Extract an internal "classic" Go function that returns `string, err` (where err is nil if no error, as usual). No CString involved.
2. The exported `evaluteGo` function will call the internal one, and then format the single C-compatible return value, including the unescaping of `\\u003e` etc.
3. I suggest the return CString to start with `Error:: ....` or `Ok:: ...` so it's more explicit.

lib/src/core/verifier/policy/opa/opa.go
```diff
@@ -0,0 +1,78 @@
...
+	decisionMap["parseInfo"] = parseInfo
+	decisionMap["allow"] = dataOPA["allow"]
+
+	decision, err := json.Marshal(decisionMap)
+	if err != nil {
+		return C.CString("Error:: " + err.Error())
+	}
+	decision_str := string(decision)
+	res := strings.Replace(decision_str, "\\u003e", ">", -1)
+	res = strings.Replace(res, "\\u003c", "<", -1)
```
dubek, line 73: instead of this post-processing to remove quoting, I suggest disabling it in the JSON encoder. something like this should work:

```go
var buf bytes.Buffer
enc := json.NewEncoder(&buf)
enc.SetEscapeHTML(false)
err := enc.Encode(decisionMap)
if err != nil {
	return C.CString("Error:: " + err.Error())
}
return C.CString(buf.String())
```

(though I think `SetEscapeHTML` is from Go 1.17, not sure if that's OK with us.)

lib/src/core/verifier/policy/opa/opa.go
```diff
@@ -0,0 +1,78 @@
...
+	parseInfo := make(map[string]interface{})
+
+	for k, v := range inputOPA {
+		value := [2]interface{}{v, data_map[k]}
+		parseInfo[k] = value
+	}
+
+	decisionMap := make(map[string]interface{})
+	decisionMap["parseInfo"] = parseInfo
+	decisionMap["allow"] = dataOPA["allow"]
```
dubek, line 65: This can be a go struct (with `json` annotations if needed) instead of a free-form map of strings.

lib/src/core/verifier/policy/opa/opa.go
```diff
@@ -0,0 +1,78 @@
...
+
+	// Make opa query
+	rs, err := query.Eval(ctx, rego.EvalInput(input_map))
+	if err != nil {
+		return C.CString("Error:: " + err.Error())
+	}
+
+	// Transform the processed decision into the format rust hopes for
+	inputOPA := rs[0].Expressions[0].Value.(map[string]interface{})
+	dataOPA := rs[0].Expressions[1].Value.(map[string]interface{})
```
dubek, line 55: Can't these type conversions panic? 
Maybe we should add another `, err` return value and check it and return error nicely to Rust?

lib/src/core/verifier/policy/opa/opa.go
```diff
@@ -0,0 +1,78 @@
...
+	input_map := make(map[string]interface{})
+	err := json.Unmarshal([]byte(input), &input_map)
+	if err != nil {
+		return C.CString("Error:: " + err.Error())
+	}
+
+	data_map := make(map[string]interface{})
+	err2 := json.Unmarshal([]byte(data), &data_map)
+	if err2 != nil {
+		// Handle error.
```
dubek, line 26: Remove the comment

lib/src/core/verifier/policy/opa/opa.go
```diff
@@ -0,0 +1,78 @@
...
+		return C.CString("Error:: " + err.Error())
+	}
+	// Manually create the storage layer. inmem.NewFromObject returns an
+	// in-memory store containing the supplied data.
+	store := inmem.NewFromObject(data_map)
+
+	// Construct a Rego object that can be prepared or evaluated.
+	r := rego.New(
+		rego.Query("input;data.policy"),
+		rego.Module("demo.rego", policy),
```
dubek, line 36: what's the meaning of `demo` here?

lib/src/core/verifier/policy/opa/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+    // Call the function exported by cgo and process the returned decision
+    let decision_buf: *mut c_char = unsafe { evaluateGo(policy_go, reference_go, input_go) };
+    let decision_str: &CStr = unsafe { CStr::from_ptr(decision_buf) };
+    let res = decision_str.to_str()?.to_string();
+    debug!("Evaluated:\n{}", res);
+    match res.starts_with("Error::") {
+        true => Err(anyhow!(res)),
+        false => Ok(res),
+    }
```
dubek, line 43: Consider: replace `match` with `if`

lib/src/core/verifier/policy/opa/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+pub struct GoString {
+    pub p: *const c_char,
+    pub n: isize,
+}
+
+pub fn evaluate(policy: String, reference: String, input: String) -> Result<String> {
+    let policy_go = GoString {
+        p: policy.as_ptr() as *const i8,
+        n: policy.len() as isize,
+    };
```
dubek, line 23: consider adding a `from_str(s: String)` function in `impl GoString` with this functionality.

lib/src/core/mod.rs
liangzhou121: Yes, I think it's good idea to pass the entire attestation payload to the AS. And the `additional-params`  can be put into the TEE specific `tee-evicence` JSON string. And jiale has added a new "tee" to designate the Evidence's type, as:
```JSON
{
    "evidence": {
        "nonce": "",
         /* TEE type name */
	"tee": "",
	"tee-pubkey": {
            "algorithm": "",
            "pubkey-length": "",
            "pubkey": ""
        },
	"tee-evidence": {}
    }
}
```

lib/src/core/mod.rs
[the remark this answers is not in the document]
liangzhou121: OK, I will splict it to a `policy_info` and a `result` fields.

lib/src/lib.rs
```diff
@@ -0,0 +1,389 @@
...
+    /// #[tokio::main]
+    /// async fn main() {
+    ///     let workdir = Path::new("./");
+    ///     let user_id = "tester".to_string();
+    ///     let service = Service::new(&workdir, user_id).await.unwrap();
+    ///     let res = service.attestation(&evidence()).await;
+    ///     assert!(res.is_ok());
+    /// }
+    /// ```
+    pub async fn attestation(&self, evidence: &String) -> Result<String> {
```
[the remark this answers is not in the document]
liangzhou121: Hi @dubek ,
I think It's better to let the invoker to convert this results to protobuf message and send it back to client. Because this `lib ` sub-project is a Crate. 
And I will add another project names `server` application which will depend on this Crate. It can response the protobuf format results.

lib/src/lib.rs
```diff
@@ -0,0 +1,389 @@
...
+    /// #[tokio::main]
+    /// async fn main() {
+    ///     let workdir = Path::new("./");
+    ///     let user_id = "tester".to_string();
+    ///     let service = Service::new(&workdir, user_id).await;
+    ///     assert!(service.is_ok());
+    /// }
+    /// ```
+    pub async fn new(path: &Path, user_id: String) -> Result<Self> {
+        let dir = path.join(user_id);
```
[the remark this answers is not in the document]
liangzhou121: I will remove this mechanism in the latest commitment. I think it isn't needed.

lib/src/fs.rs
```diff
@@ -0,0 +1,28 @@
...
+        if bak.exists() {
+            // Copy back
+            fs::copy(&bak, &src).await?;
+        }
+        return Err(anyhow!("error: set file failed."));
+    }
+    if bak.exists() {
+        fs::remove_file(&bak).await?;
+    }
+    Ok(())
```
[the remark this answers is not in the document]
liangzhou121: I will remove the disk access codes totally in the last commitment. I think it's not a good a idea that a `Crate` stores some persistent files/state in the filesystem.

lib/src/core/verifier/policy/opa/opa.go
liangzhou121: Actually I don't familir with the golang. Do you have any suggestions about how to handle the potential error gracefully here? Thank you!

dubek: See [this example](https://go.dev/tour/methods/15). I think in this case it should be:

```suggestion
	inputOPA, ok := rs[0].Expressions[0].Value.(map[string]interface{})
	if !ok {
	      retrun C.CString("Error:: unexpected type in first expression")
	}
	dataOPA, ok := rs[0].Expressions[1].Value.(map[string]interface{})
	if !ok {
	      retrun C.CString("Error:: unexpected type in second expression")
	}
```

lib/src/lib.rs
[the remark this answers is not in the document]
dubek: I agree that the lib (crate) should not return protobuf messages. But maybe it can returns a Rust struct with meaningful fields instead of String -- if it makes sense (that is, if there are several fixed fields with known meaning/use).

lib/src/fs.rs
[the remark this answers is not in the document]
dubek: Sounds good.

lib/src/core/verifier/policy/opa/opa.go
liangzhou121: Thank you very much!
But it will cause the following error. So I will keep the current implemenation:

  ERROR: # command-line-arguments
  ./opa.go:53:16: cannot assign bool to err (type error) in multiple assignment:
        bool does not implement error (missing Error method)
  ./opa.go:57:15: cannot assign bool to err (type error) in multiple assignment:
        bool does not implement error (missing Error method)
  note: module requires Go 1.16

lib/src/core/verifier/policy/opa/mod.rs
liangzhou121: I tried to add the following codes, but it will cause execution failure. It is weired.
```
fn from_str(s: String) -> GoString {
    GoString {
        p: s.as_ptr() as *const i8,
        n: s.len() as isize,
    }  
}
```

lib/src/core/verifier/policy/opa/opa.go
dubek: If you've used `err` instead of `ok` as the second return variable this will not work, because `err` is an already-defined variable of type `error` (whereas type assertions return bool as the second return value).

lib/src/core/verifier/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+    fn opa_policy_path(&self) -> Result<PathBuf>;
+    fn opa_reference_data_path(&self) -> Result<PathBuf>;
+}
+
+#[derive(Debug)]
+pub struct Verifiers {
+    pub tdx: tdx::Tdx,
+    pub sgx: sgx::Sgx,
+    pub sample: sample::Sample,
+}
```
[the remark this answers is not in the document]
liangzhou121: Yes, I will reconstruct the codes according to your comments. And I also rethink the `evaluation()` and concluded that let this Crate to manage and store some persistent files(such as policy and reference data) is inapporiate. So I switch the function interface into this:
```RUST
    async fn evaluate(
        &self,
        evidence: &Evidence,
        policy: Option<String>,
        reference_data: Option<String>,
    ) -> Result<AttestationResults> ;
```
So I it can receive the external policy and reference data to execute evaluation. If the parameters are empty, it will use the   default policy and reference data.

lib/src/lib.rs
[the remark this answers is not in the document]
liangzhou121: Yes, this string is a JSON string and it is Deserialized by this structure:
```RUST
#[derive(Serialize, Deserialize, Debug)]
pub struct AttestationResults {
    pub tee: String,     // The TEE type, such as  "tdx", "sev-snp"
    pub result: String,   // The policy engine's evaluation reults.
    pub policy_info: String,   // It's also a JSON string, you can check the \src\core\verifier\sample\mod.rs for more info
    pub tcb: String,   // It's also a JSON string, please check the \src\core\verifier\sample\mod.rs for more info
}
```

lib/src/core/verifier/policy/opa/opa.go
liangzhou121: I find this `Replace` isn't needed here, so I will remove this snippet codes.

lib/src/core/verifier/policy/opa/opa.go
liangzhou121: Yes, use a struct will make the codes clear, but this glue code is straightforward, so I think we needn't add too much optimization here.

liangzhou121: Yes, you comment reminds me that a bug exist here. This code needs to pass all the whole OPA evaluation results to the RUST, and the evaluation is dynamically according to the `Policy` file. So it shouldn't use a go struct here but I will update the codes as the following:
```
	// OPA's evaluation results.
	for k, v := range dataOPA {
	    decisionMap[k] = v
	}
	// The input and reference data pair.
	decisionMap["parseInfo"] = parseInfo
```
And I will also enhance the `core\verifier\policy\opa\mod.rs` cargo test cases to cover this scenario.

lib/src/core/verifier/mod.rs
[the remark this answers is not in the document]
sameo: Please let me know when the updated implementation is ready for a follow up review.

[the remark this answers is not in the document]
liangzhou121: Yes, I have updated this PR according to all these review comments. And I also found and fixed two bugs during some development working which depends on this PR's Crate.
@sameo @fitzthum @dubek Please help to review this PR again.

lib/src/lib.rs
```diff
@@ -0,0 +1,340 @@
...
+    /// }
+    /// "#.to_string();
+    ///
+    ///     let reference_data = json!({
+    ///        "cpusvn": 1,
+    ///        "svn": 1
+    ///     }).to_string();
+    ///
+    ///     // Attest the evidence with customized OPA policy and reference data.
+    ///     let res = service.attestation(&evidence(), Some(policy), Some(reference_data)).await;
```
sameo, line 109: That looks much better now. We get an evaluation based on an evidence, a policy and some reference data.

sameo (commented):
I just have a couple of code organization comments left to address. Otherwise, this code looks nice to me, thanks @liangzhou121

lib/src/core/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+        policy: Option<String>,
+        reference_data: Option<String>,
+    ) -> Result<String> {
+        let evidence =
+            serde_json::from_str::<Evidence>(evidence).context("Deserialize Evidence failed.")?;
+        let tee = match evidence.tee.as_str() {
+            "sample" => Ok(Box::new(sample::Sample::default()) as Box<dyn Verifier + Send>),
+            "sgx" => Ok(Box::new(sgx::Sgx::default()) as Box<dyn Verifier + Send>),
+            "tdx" => Ok(Box::new(tdx::Tdx::default()) as Box<dyn Verifier + Send>),
+            _ => Err(anyhow!("not supported!")),
```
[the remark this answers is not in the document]
liangzhou121: Yes, Your suggestion is really a good approach. I updated the codes according to it.

lib/src/lib.rs
liangzhou121: Yes, the `attesation()` can receivce policy and reference data file dynamically now. If the policy/reference data file is None, it will use the default file.

jodh-intel (commented):
Thanks @liangzhou121 - a few comments...

lib/src/core/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+        let results = serde_json::to_string(&results)?;
+        debug!("Attestation Results:\n{}", &results);
+        Ok(results)
+    }
+
+    pub fn get(&self, tee: TEE, file: File) -> Result<String> {
+        let tee = match tee {
+            TEE::SGX => Ok(Box::new(sample::Sample::default()) as Box<dyn Verifier>),
+            TEE::TDX => Ok(Box::new(sgx::Sgx::default()) as Box<dyn Verifier>),
+            TEE::SAMPLE => Ok(Box::new(tdx::Tdx::default()) as Box<dyn Verifier>),
```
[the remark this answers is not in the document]
jodh-intel: Ah - this now seems to have been fixed.

lib/src/core/verifier/policy/opa/opa.go
```diff
@@ -0,0 +1,79 @@
+package main
+
+import "C"
+
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/open-policy-agent/opa/rego"
```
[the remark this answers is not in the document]
liangzhou121: There is also another private project named [opa-rs](https://github.com/myagley/opa-rs). But it's also unoffical. So I prefer to implement the shim codes by ourselves now. 
May be we can switch some opa-rs Crate if in future.

lib/src/core/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+pub enum File {
+    Policy,
+    Reference,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Evidence {
+    pub nonce: String,
+    pub tee: String,
+    #[serde(rename = "tee-pubkey")]
```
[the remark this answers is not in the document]
liangzhou121: I googled and didn't find some key word like `rename_pattern`.
Actually the current implement have already designated the the rename pattern explictly. As the following:
```
    #[serde(rename = "tee-pubkey")]
    pub tee_pubkey: String,
```
It means rename the input's `tee-pubkey` -> `tee_pubkey`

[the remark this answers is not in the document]
jodh-intel: Sure. I wasn't expecting you to do this, but I'm saying it would be great if `serde` allowed you to apply the pattern to the *type*, not the field. That way, you could specify a single pattern potentially, rather than one for every single field in the type.

lib/src/core/verifier/policy/opa/opa.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/open-policy-agent/opa/storage/inmem"
+)
+
+//export evaluateGo
+func evaluateGo(policy string, data string, input string) *C.char {
```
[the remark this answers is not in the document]
liangzhou121: It should be the RUST's limitation, The RUST is compatible with the C language ABI. So this `evaluateGo()` function's return value is also converted to `C.char` type.

lib/src/core/verifier/sample/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+
+#[derive(Serialize, Deserialize, Debug)]
+struct Ehd {
+    nonce: String,
+    public_key: String,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+struct Tcb {
+    is_debuggable: bool,
```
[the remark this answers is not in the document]
liangzhou121: It's only a `sample` to demonstrate the potential TEE's TCB status. Take SGX TEE as example, it means whether the SGX enclave program can be debugged.

lib/src/core/verifier/sample/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+    async fn evaluate(
+        &self,
+        evidence: &Evidence,
+        policy: Option<String>,
+        reference_data: Option<String>,
+    ) -> Result<AttestationResults> {
+        // Use the default policy/reference_data if the input is None.
+        let policy = policy.unwrap_or(std::include_str!("policy.rego").to_string());
+        let reference_data =
+            reference_data.unwrap_or(std::include_str!("reference_data.json").to_string());
```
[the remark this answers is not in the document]
liangzhou121: I will define them as const global values.

lib/src/core/verifier/sgx/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+#[derive(Serialize, Deserialize, Debug)]
+struct Ehd {
+    nonce: String,
+    public_key: String,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+struct Tcb {
+    collateral: Collateral,
+    ehd: Ehd,
```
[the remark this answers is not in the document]
liangzhou121: Ehd == Enclave Hold Data

lib/src/lib.rs
```diff
@@ -0,0 +1,345 @@
...
+mod core;
+use crate::core::verifier::policy::opa;
+
+/// The supported TEE types:
+/// - TDX: TDX TEE.
+/// - SGX: SGX TEE.
+/// - SEVSNP: SEV-SNP TEE.
+/// - SAMPLE: A dummy TEE that used to test/demo the attestation service functionalities.
+#[derive(Debug, EnumString)]
+pub enum TEE {
```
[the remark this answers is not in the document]
liangzhou121: Because the TEE enum needs to be exported by this Crate:
```
use attestation_service::{Service, TEE};

#[tokio::main]
async fn main() {
    let service = Service::new().unwrap();
    let res = service.default_policy(TEE::SAMPLE).await;
    assert!(res.is_ok());
 }
```
But the TEE enum's implementation is internal.

lib/src/core/mod.rs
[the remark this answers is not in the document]
liangzhou121: OK, But seems the [Serde](https://serde.rs/field-attrs.html) doesn't support this rename/replace approch.

lib/src/core/mod.rs
```diff
@@ -0,0 +1,78 @@
...
+#[derive(Serialize, Deserialize, Debug)]
+pub struct AttestationResults {
+    pub tee: String,
+    pub result: String,
+    pub policy_info: String,
+    pub tcb: String,
+}
+
+impl TEE {
+    pub fn from_str(str: &str) -> Result<Self> {
```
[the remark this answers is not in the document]
liangzhou121: I think I prefer to keep this implementation and the `from_str()` also align with the naming of `FromStr` trait.

lib/src/core/verifier/sgx/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+
+#[derive(Serialize, Deserialize, Debug)]
+struct Collateral {
+    qeidcertshash: String,
+    qeidcrlhash: String,
+    qeidhash: String,
+    quotehash: String,
+    tcbinfocertshash: String,
+    tcbinfocrlhash: String,
+    tcbinfohash: String,
```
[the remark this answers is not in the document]
liangzhou121: I will delete this structures in this PR. They will be added back when we add the SGX's supportion in future. This should be better.

fitzthum (approved):
I think we will still need to make some adjustments, particularly when we try to add in the app code and implement the different verifier modules, but I think this is close enough to move on.

sameo (approved):
Thanks a lot @liangzhou121, this LGTM now.

sameo:
@jodh-intel Could you please take another look?

jodh-intel (approved):
Thanks @liangzhou121.

lgtm

PR #4, opened 2022-06-06
AS: Add a server application

liangzhou121:
1.Implemented the Attestation Service's server application.

Fixes: #3
Signed-off-by: zhouliang121 <liang.a.zhou@linux.alibaba.com>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,5 @@
 [workspace]
 members = [
   "lib",
+  "server",
 ]
--- a/server/README.md
+++ b/server/README.md
@@ -1,0 +1,181 @@
+# Attestation Server
+
+Attestation Server is a user space application for attestation purpose. 
+It receives and verifies the TEE's attestation [Evidence](https://github.com/confidential-containers/kbs/blob/main/docs/kbs_attestation_protocol.md#attestation) and responses the corresponding [Attestation Results](https://github.com/confidential-containers/attestation-service/issues/1) to ensure the Evidence's generation environment is a real TEE environment and its Trusted Computing Base (TCB) status is as expected.
+
+Attestation Server supports the following types of TEE:
+- SGX
+- TDX
+- SEV-SNP
+- SAMPLE: Dummy TEE which is used to test/demo the Server's functionalities.
+
+Attestation Server depends on [Open Policy Agent (OPA)](https://www.openpolicyagent.org/docs/latest/) to evaluate the Evidence's TCB status during attestation. Each supported TEE's OPA `Policy(.rego)` and `Reference Data(.json)` can be customized in order to evaluate the TCB status precisely.
+Note: Please refer [Policy Language](https://www.openpolicyagent.org/docs/latest/policy-language/) for more about `.rego` syntax information.
+
+## API
+
+Attestation Server provides two groups of GRPC endpoints:
+- attestation: Attest the received Evidence which is defined in [protobuf](https://github.com/confidential-containers/attestation-service/server/proto/attestation.proto).
+- management: Customize the Server's configuration which is defined in [protobuf](https://github.com/confidential-containers/attestation-service/server/proto/management.proto).
+
+### Attestation
+
+The `attestation` group only includes one `Attestation` GRPC endpoint currently. It receives the attestation [Evidence](https://github.com/confidential-containers/kbs/blob/main/docs/kbs_attestation_protocol.md#attestation), executes attestation procedures, and and responds with the [Attestation Results](https://github.com/confidential-containers/attestation-service/issues/1).
+
+`Attestation` message:
+```PROTO
+message AttestationRequest {
+    // Attestation evidence: https://github.com/confidential-containers/kbs/blob/main/docs/kbs_attestation_protocol.md#attestatio
+    bytes evidence = 1;
+    // Optional: Designate the user id. It should be kept as "None" currently.
+    optional string user = 2;
+}
+```
+
+### Management
+
+It's mainly used to customize Attestation Server's configurations:
+- Customize each type of TEE's `Policy(.rego)` and `Reference Data(.json)` in order to evaluate it's TCB status precisely.
+    - GetPolicy
+    - GetReferenceData
+    - SetPolicy
+    - SetReferenceData
+    - RestoreDefaultPolicy
+    - RestoreDefaultReferenceData
+- Test/Evaluate the local `Policy(.rego)` or `Reference Data(.json)` with remote Attestation Server's OPA engine.
+    - TestOpa
+
+The supported types of TEE:
+```PROTO
+enum Tee {
+    Sgx = 0;
+    Tdx = 1;
+    SevSnp = 2;
+    Sample = 3;
+}
+```
+
+#### GetPolicy
+
+Get the specific TEE's OPA `Policy(.rego)`. The message:
+```PROTO
+message GetPolicyRequest {
+    // Supported TEE types
+    Tee tee = 1;
+    // Optional: Designate the user id. It should be kept as "None" currently.
+    optional string user = 2;
+}
+```
+
+#### GetReferenceData
+
+Get the specific TEE's OPA `Reference Data(.json)`. The message:
+```PROTO
+message GetReferenceDataRequest {
+    // Supported TEE types
+    Tee tee = 1;
+    // Optional: Designate the user id. It should be kept as "None" currently.
+    optional string user = 2;
+}
+```
+
+#### SetPolicy
+
+Set the specific TEE's OPA `Policy(.rego)`. It can make the Attestation Server evaluate the TEE TCB status according to specific user's preference. And it will echo error if the new `Policy(.rego)` syntax is illegal. The message:
+```PROTO
+message SetPolicyRequest {
+    // Supported TEE types
+    Tee tee = 1;
+    // Optional: Designate the user id. It should be kept as "None" currently.
+    optional string user = 2;
+    // The "Policy(.rego)" file's content.
+    bytes content = 3;
+}
+```
+
+#### SetReferenceData
+
+Set the specific TEE's OPA `Reference Data(.json)`. It can attach a new released program's reference measurement value into the corresponding program's allow list. The message:
+```PROTO
+message SetReferenceDataRequest {
+    // Supported TEE types
+    Tee tee = 1;
+    // Optional: Designate the user id. It should be kept as "None" currently.
+    optional string user = 2;
+    // The "Reference Data(.json)" file's content.
+    bytes content = 3;
+}
+```
+
+#### RestoreDefaultPolicy
+
+Restore the specific TEE's OPA `Policy(.rego)` to default value. The message:
+```PROTO
+message ResetPolicyRequest {
+    // Supported TEE types
+    Tee tee = 1;
+    // Optional: Designate the user id. It should be kept as "None" currently.
+    optional string user = 2;
+}
+```
+
+#### RestoreDefaultReferenceData
+
+Restore the specific TEE's OPA `Reference Data(.json)` to default value. The message:
+```PROTO
+message ResetReferenceDataRequest {
+    // Supported TEE types
+    Tee tee = 1;
+    // Optional: Designate the user id. It should be kept as "None" currently.
+    optional string user = 2;
+}
+```
+
+#### TestOpa
+
+It's main purpose is to ensure the updated `Policy(.rego)` or `Reference Data(.json)` can work as expected before upload to remote Attestation Server. The message:
+```PROTO
+message TestOpaRequest {
+    // The OPA "Policy(.rego)" content that need to test.
+    bytes policy = 1;
+    // The "Reference Data(.json)" content that need to test.
+    bytes reference = 2;
+    // The "input" which will be evaluated by the request's "policy" and "referece".
+    bytes input = 3;
+}
+```
+
+## Usage
+
+Here are the steps of building and running Attestation Server:
+
+### Build
+
+Build Attestation Server:
+```shell
+git clone https://github.com/confidential-containers/attestation-service
+cd attestation-service
+cargo build --release
+```
+
+### Run
+
+- For help information, run:
+```shell
+./target/release/attestation-server --help
+```
+
+- For version information, run:
+```shell
+./target/release/attestation-server --version
+```
+
+Start Attestation Server and specify the attestation and management listen ports of its gRPC service:
+```shell
+./target/release/attestation-server --attestation-sock 127.0.0.1:3000 --management-sock 127.0.0.1:3001
+```
+
+If you want to see the runtime log, run:
+```shell
+RUST_LOG=debug ./target/release/attestation-server --attestation-sock 127.0.0.1:3000 --management-sock 127.0.0.1:3001
+```
--- a/server/proto/attestation.proto
+++ b/server/proto/attestation.proto
@@ -1,0 +1,17 @@
+syntax = "proto3";
+
+import "common.proto";
+
+package attestation;
+
+message AttestationRequest {
+    bytes evidence = 1;
+    optional common.User user = 2;
+}
+message AttestationResponse {
+    bytes attestation_results = 1;
+}
+
+service AttestationService {
+    rpc Attestation(AttestationRequest) returns (AttestationResponse) {};
+}
--- a/server/proto/management.proto
+++ b/server/proto/management.proto
@@ -1,0 +1,70 @@
+syntax = "proto3";
+
+import "common.proto";
+
+package management;
+
+enum Tee {
+    Sgx = 0;
+    Tdx = 1;
+    SevSnp = 2;
+    Sample = 3;
+}
+
+message GetPolicyRequest {
+    Tee tee = 1;
+    optional common.User user = 2;
+}
+message GetPolicyResponse {
+    bytes content = 1;
+}
+
+message GetReferenceDataRequest {
+    Tee tee = 1;
+    optional common.User user = 2;
+}
+message GetReferenceDataResponse {
+    bytes content = 1;
+}
+
+message SetPolicyRequest {
+    Tee tee = 1;
+    optional common.User user = 2;
+    bytes content = 3;
+}
+message SetPolicyResponse {}
+
+message SetReferenceDataRequest {
+    Tee tee = 1;
+    optional common.User user = 2;
+    bytes content = 3;
+}
+message SetReferenceDataResponse {}
+
+message RestoreDefaultPolicyRequest {
+    Tee tee = 1;
+    optional common.User user = 2;
+}
+message RestoreDefaultPolicyResponse {}
+
+message RestoreDefaultReferenceDataRequest {
+    Tee tee = 1;
+    optional common.User user = 2;
+}
+message RestoreDefaultReferenceDataResponse {}
+
+service ManagementService {
+    // Get the GetPolicyRequest.user and GetPolicyRequest.tee specified Policy(.rego)
+    rpc GetPolicy(GetPolicyRequest) returns (GetPolicyResponse) {};
+    // Get the GetReferenceDataRequest.user and GetReferenceDataRequest.tee specified Reference Data(.json)
+    rpc GetReferenceData(GetReferenceDataRequest) returns (GetReferenceDataResponse) {};
+    // Update the SetPolicyRequest.user and SetPolicyRequest.tee specified Policy(.rego) with the content of SetPolicyRequest.content
+    rpc SetPolicy(SetPolicyRequest) returns (SetPolicyResponse) {};
+    // Update the SetReferenceDataRequest.user and SetReferenceDataRequest.tee specified Reference Data(.json) with the content of SetReferenceDataRequest.content
+    rpc SetReferenceData(SetReferenceDataRequest) returns (SetReferenceDataResponse) {};
+    // Restore the RestoreDefaultPolicyRequest.user and RestoreDefaultPolicyRequest.tee specified Policy(.rego) to default. 
+    rpc RestoreDefaultPolicy(RestoreDefaultPolicyRequest) returns (RestoreDefaultPolicyResponse) {};
+    // Restore the RestoreDefaultReferenceDataRequest.user and RestoreDefaultReferenceDataRequest.tee specified Reference Data(.json) to default. 
+    rpc RestoreDefaultReferenceData(RestoreDefaultReferenceDataRequest) returns (RestoreDefaultReferenceDataResponse) {};
+}
+
--- a/server/src/attestation.rs
+++ b/server/src/attestation.rs
@@ -1,0 +1,294 @@
+use anyhow::Result;
+use attestation_service::Service as AS;
+use std::path::PathBuf;
+use std::sync::Arc;
+use tokio::sync::RwLock;
+use tonic::{Request, Response, Status};
+extern crate serde;
+use self::serde::{Deserialize, Serialize};
+use crate::user;
+use tonic::transport::Server;
+
+use crate::attestation_api::attestation_service_server::{
+    AttestationService, AttestationServiceServer,
+};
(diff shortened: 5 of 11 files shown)


server/src/users/mod.rs
```diff
@@ -0,0 +1,11 @@
...
+
+pub mod policy;
+
+const DEFAULT_USER_ID: &str = "default";
+
+// Fetch the users working directory.
+pub fn path(dir: &Path, id: Option<String>) -> PathBuf {
+    let id = id.unwrap_or_else(|| DEFAULT_USER_ID.to_string());
+    dir.join("users").join(id)
+}
```
dubek, line 11: 1. I see that in all the call-sites of this function we add another directory element:

```
let file = users::path(dir, id).join(tee) ...
```

maybe add the tee string as another argument to this function?

2. No sanitization/validation is done on value of `id` (or `tee`) which are received from the client request. They may contain `"../../../"` and not sure what this might cause.

server/README.md
```diff
@@ -0,0 +1,181 @@
...
+```PROTO
+message AttestationRequest {
+    // Attestation evidence: https://github.com/confidential-containers/kbs/blob/main/docs/kbs_attestation_protocol.md#attestatio
+    bytes evidence = 1;
+    // Optional: Designate the user id. It should be keeped as "None" currently.
+    optional string user = 2;
+}
+```
+
+### Management
```
fitzthum, line 35: The management API is security sensitive. Anyone who can access these endpoints can control the AS. How are we going to make sure that only a trusted party can manage the AS? Are these commands going to be routed through the KBS or come directly from the guest owner?

server/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+### Attestation
+
+The `attestation` group only includes one `Attestation` GRPC endpoint currently. It receives the attestation [Evidence](https://github.com/confidential-containers/kbs/blob/main/docs/kbs_attestation_protocol.md#attestation), executes attestation procedures, and responses the [Attestation Results](https://github.com/confidential-containers/attestation-service/issues/1).
+
+`Attestation` message:
+```PROTO
+message AttestationRequest {
+    // Attestation evidence: https://github.com/confidential-containers/kbs/blob/main/docs/kbs_attestation_protocol.md#attestatio
+    bytes evidence = 1;
```
fitzthum, line 29: This is the full evidence message (including the nonce and public key) and not just the `tee-evidence` field, right?

server/README.md
```diff
@@ -0,0 +1,181 @@
...
+    // Supported TEE types
+    Tee tee = 1;
+    // Optional: Designate the user id. It should be keeped as "None" currently.
+    optional string user = 2;
+}
+```
+
+#### SetPolicy
+
+Set the specific TEE's OPA `Policy(.rego)`. It can make the Attestation Server evaluate the TEE TCB status according to specific user's preference. And it will echo error if the new `Policy(.rego)` syntax is illegal. The message:
```
fitzthum, line 84: So we're going to parse each policy file when it is uploaded to make sure the syntax is correct? Is only one policy allowed per tee? Does uploading a new one overwrite an old one?

server/README.md
```diff
@@ -0,0 +1,181 @@
...
+    // Optional: Designate the user id. It should be keeped as "None" currently.
+    optional string user = 2;
+    // The "Policy(.rego)" file's content.
+    bytes content = 3;
+}
+```
+
+#### SetReferenceData
+
+Set the specific TEE's OPA `Reference Data(.json)`. It can attach a new released program's reference measurement value into the corresponding program's allow list. The message:
```
fitzthum, line 98: Only one reference file allowed per tee? Should we also provide the name of the reference data?

server/README.md
```diff
@@ -0,0 +1,181 @@
...
+    // Optional: Designate the user id. It should be keeped as "None" currently.
+    optional string user = 2;
+    // The "Reference Data(.json)" file's content.
+    bytes content = 3;
+}
+```
+
+#### ResetPolicy
+
+Reset the specific TEE's OPA `Policy(.rego)` to default value. The message:
```
fitzthum, line 112: It might make more sense to call this `DeletePolicy`. We aren't updating the policy to make it the same as the reference data, we are just removing the policy and then using the default policy instead. Same for reference data.

server/README.md
```diff
@@ -0,0 +1,181 @@
...
+It's mainly used to customize Attestation Server's configurations:
+- Customize each type of TEE's `Policy(.rego)` and `Reference Data(.json)` in order to evaluate it's TCB status precisely.
+    - GetPolicy
+    - GetReferenceData
+    - SetPolicy
+    - SetReferenceData
+    - ResetPolicy
+    - ResetReferenceData
+- Test/Evaluate the local `Policy(.rego)` or `Reference Data(.json)` with remote Attestation Server's OPA engine.
+    - TestOpa
```
fitzthum, line 46: Should we also have an endpoint to get logging information from the AS or is that better left to the KBS?

fitzthum (commented):
Just a few high-level comments for now. I'll try to dig into the code soon.

lib/src/core/verifier/policy/opa/mod.rs
```diff
@@ -102,34 +102,27 @@ svn_old_allow {
...
-            dummy_input(5, 5)
-        );
+        let res = evaluate(policy.clone(), dummy_reference(3), dummy_input(5, 5));
         assert!(res.is_ok(), "OPA execution() should success");
         let v: Value = serde_json::from_str(&res.unwrap()).unwrap();
         assert!(v["allow"] == true, "allow should true");
-        assert!(v["product_id_allow"] == true, "product_id_allow should true");
+        assert!(
+            v["product_id_allow"] == true,
+            "product_id_allow should true"
```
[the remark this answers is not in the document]
liangzhou121: I made a mistake because these changes should not be included in this PR. I will revert it. Or it will conflict with the "Add basic compile testing" PR.

server/README.md
liangzhou121: Yes, this is the full evidence message. I followed your previous suggestion that KBS should relay the evidence payload to AS directly. And the `tee-evidence` will only be parsed inside attestation service Crate internally.

server/README.md
liangzhou121: > The management API is security sensitive. Anyone who can access these endpoints can control the AS. How are we going to make sure that only a trusted party can manage the AS?

In POC stage, the Attestation Server should run in the same trusted environment as the potential management tool and the KBS. So the Attestation Server must open a local GRPC address (127.0.0.1:1234) to accept attestation request and management requests.

> Are these commands going to be routed through the KBS or come directly from the guest owner?

The Server will provide independent management GRPC endpoints. So it can receive request from both kbs and guest owner.

server/README.md
liangzhou121: Acutally I didn't consider to enable the logging information's storage for AS now. Currently, AS's debug log can be enabled by `RUST_LOG=debug attestation-server xxxxx`. Maybe we can add a better logging mechanism and extend the endpoint set to add a `get_logging()` endpoint in other PR in future.

server/README.md
liangzhou121: > So we're going to parse each policy file when it is uploaded to make sure the syntax is correct? 

Yes, I think the policy file's syntax should be checked, or it will make the evaluation unworkable.

> Is only one policy allowed per tee? Does uploading a new one overwrite an old one?

Yes, the current implementation is one policy per TEE. 
Yes, the upload new one will cover the old one directly and the old one won't be stored by AS (except the default policy.rego because it is intergated inside AS crate.)

server/README.md
liangzhou121: Currently it only supports one reference file per TEE, so the reference data file name is hardcoded as `reference_data.json`.
Do you think it needs to support multiple reference data files to a same TEE?

server/README.md
liangzhou121: Actually I also named this endpoint as `DeletePolicy`. But later I think `ResetPolicy` should be better. 
Because the `DeletePolicy` can only delete the user customized policy file, it can't delete the default policy file which is integrated inside AS Crate. After the customized policy file is deleted:
- The default policy will be selected and used during evidence attestation.
- `GetPolicy` will also return the default policy file.

server/src/users/mod.rs
liangzhou121: > maybe add the tee string as another argument to this function?

OK. I will add a new function which will includes the `tee` argument in `policy.rs`, as the following:
```
fn workdir(dir: &Path, id: Option<String>, tee: String) -> PathBuf {
    users::path(dir, id).join(tee)
}
```

> No sanitization/validation is done on value of id (or tee) which are received from the client request. They may contain "../../../" and not sure what this might cause

- The `tee` is restrict by `management_api::Tee` enum. So the only possible input is "sgx", "tdx", "sevsnp", and "sample".
- The `id` needs to be kept as `None` because the user id is not supported by guest client currently. Maybe we can add that checking when the user id is enabled in future.

server/README.md
fitzthum: I guess that's fine. It might be nice to be able to store multiple policies even if only one is active. That way you could easily switch between policies without having to upload them again. We can add that later, though.

server/README.md
fitzthum: It might be nice given that the reference data can be somewhat large. For SNP, for instance we're going to need to have the reference data include the entire firmware binary (not just the hash). It might be nice if we could upload this separately from the other pieces. Again, we can probably do this later.

server/README.md
fitzthum: Maybe `RestoreDefaultPolicy`?

server/README.md
liangzhou121: Yes, I agree with you that maybe it's better to let KBS keep this recording. For AS, the Attesattion Results includes all these evaluation information, so I think it should be enough to let the results act as logging also.

server/README.md
liangzhou121: OK, I think we can extend the interface to support this during such as SNP's implementation. For TDX scenario, I think one reference data file should be enough.

server/README.md
liangzhou121: I also renamed the `ResetReferenceData()` -> `RestoreDefaultReferenceData()`

server/src/main.rs
```diff
@@ -0,0 +1,55 @@
...
+            Arg::with_name("listen")
+                .short("l")
+                .long("listen")
+                .value_name("sockaddr")
+                .help("Socket that the server will listen on to accept attestation & configuration requests.")
+                .takes_value(true),
+        )
+        .get_matches();
+
+    let sockaddr = match matches.is_present("listen") {
```
fitzthum, line 45: Maybe try something along the lines of `let sockaddr = matches.value_of("listen").unwrap_or("default".to_string());`

server/src/main.rs
```diff
@@ -0,0 +1,55 @@
...
+        true => matches.value_of("listen").unwrap().to_string(),
+        false => "127.0.0.1:1234".to_string(),
+    };
+
+    debug!("Server listen addr: {}", &sockaddr);
+    let res = server::server(&sockaddr).await;
```
fitzthum, line 51: So we're going to serve both the management and attestation api on the same port?

server/src/server.rs
```diff
@@ -0,0 +1,29 @@
...
+use crate::attestation_api::attestation_service_server::AttestationServiceServer;
+use crate::management;
+use crate::management_api::management_service_server::ManagementServiceServer;
+use anyhow::Result;
+use std::path::Path;
+use std::sync::Arc;
+use tokio::sync::RwLock;
+use tonic::transport::Server;
+
+const ATTESTATION_SEVER_WORKDIR: &str = "/opt/attestation-server";
```
fitzthum, line 11: `ATTESTATION_SEVER_WORKDIR` -> `ATTESTATION_SERVER_WORKDIR`

fitzthum (commented):
This looks pretty good. I will finish review next week.

server/src/main.rs
liangzhou121: Yes, I think it's better to keep the server's launch arguments as simple as possible. So a potential user needn't to distingush each port's functionalities. And for the potential security problem (because the managment API is sensitive), we can handle it like this:
- If the connection is built with a `normal` account, the server will only accept attestation request.
- If the connection is built with a `admin` account, the server will accept both the attestation and configuration requests.

fitzthum: The attestation agent uses two different ports or the get resource and keyprovider APIs. I think that makes sense when the endpoints are used by different clients. I don't have a strong opinion here, but I think two might be better. It seems like the authentication for different accounts isn't implemented yet. If we use two different ports we could expose the attestation port to the network while keeping the other port hidden. I guess in the short term everything will be local anyway since the AS and KBS will be on the same node.

liangzhou121: Yes, It's reasonable to reach to better performance by hide the management port explictly. I will splict it to two ports.

server/src/management.rs
```diff
@@ -0,0 +1,573 @@
...
+        let request: SetPolicyRequest = request.into_inner();
+        let policy = std::str::from_utf8(&request.content)
+            .map_err(|e| Status::invalid_argument(format!("Parse content: {}", e)))?
+            .to_owned();
+        debug!("Policy: {}", &policy);
+
+        // Check new policy's syntax
+        AS::new()
+            .map_err(|e| Status::aborted(format!("Create AS: {}", e)))?
+            .opa_test(policy.clone(), "{}".to_string(), "{}".to_string())
```
fitzthum, line 84: so if we call `opa_test` with no input or reference only the syntax of the policy will be checked. It might be good to note somewhere that this won't check if the reference data matches what is expected by the policy. 

I don't think there is any way that this could be exploited with a blank input and reference, but it makes me slightly nervous. Probably fine.

fitzthum (approved):
A few more things but I think this is mostly good. Let's try to find some more people to review.

server/src/management.rs
liangzhou121: I think it should be fine, take the following Policy.rego file as example:
```
package policy

# By default, deny requests. 
default allow = false

allow {
    input.cpusvn >= data.cpusvn
    input.svn >= data.svn
}
```
If the `input.json` and `reference_data.json` files are empty, the evaluation result is: `{"allow":false,"parseInfo":{}}`
It shows the `allow == false` explictly.

But your concern reminds us that a new `Policy.rego` should be tested with empty `input.json` and `reference_data.json` to ensure it can work as expected also.

server/src/main.rs
```diff
@@ -0,0 +1,72 @@
...
+
+#[macro_use]
+extern crate log;
+shadow!(build);
+
+mod attestation;
+mod management;
+mod users;
+
+const ATTESTATION_SERVER_WORKDIR: &str = "/opt/attestation-server";
```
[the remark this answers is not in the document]
liangzhou121: OK, we can add it as command line argument option in future. 
If the argument is empty, then this `/opt/attestation-server` default path will be used.

server/src/attestation.rs
```diff
@@ -0,0 +1,257 @@
...
+// The evidence is deserializes by Attestation Server partially, and it will be deserialized wholly by Attesattion Service Crate.
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Evidence {
+    // The evidence's TEE type.
+    pub tee: String,
+}
+
+#[derive(Debug)]
+pub struct Service {
+    locker: Arc<RwLock<u8>>,
```
sameo, line 30: Rust locking APIs are designed to associate a  lock/mutex with the data it protects. What you're doing here is closer to a C-ish approach, and I think this calls for a slightly different pattern that would look like that:

```Rust
struct InnerService {
     workdir: PathBuf,
}

pub struct Service {
     service: Arc<RwLock<InnerService>>,
}

impl Service {
    pub fn new(workdir: PathBuf) -> Self {
        Self {
            service: Arc::new(RwLock::new(InnerService { workdir })),
        }
    }
}
```

With that pattern the lock becomes transparent for anyone using `Service`, and you have the guarantee that one must take the lock if it wants to use the actual `Service` data, even when/if we expand the `Service` API or structure in the future.

Also callers of the API don't need to create an artificial lock, that's handled internally.

lib/src/core/mod.rs
```diff
@@ -22,6 +22,8 @@ macro_rules! default_reference_data {
     };
 }
 
+// The evidence(.json) must align with the "attestation" definition in:
+// https://github.com/confidential-containers/kbs/blob/main/docs/kbs_attestation_protocol.md#attestation
```
sameo, line 26: I dont think this change should be part of this PR.

server/src/management.rs
```diff
@@ -0,0 +1,571 @@
...
+            Tee::SevSnp => "sevsnp",
+            Tee::Sample => "sample",
+        };
+        write!(f, "{}", str)
+    }
+}
+
+#[derive(Debug)]
+pub struct Service {
+    locker: Arc<RwLock<u8>>,
```
sameo, line 52: Same comment as the one from `attestation.rs` applies here.

server/src/management.rs
```diff
@@ -0,0 +1,571 @@
...
+use crate::management_api::{
+    GetPolicyRequest, GetPolicyResponse, GetReferenceDataRequest, GetReferenceDataResponse,
+    RestoreDefaultPolicyRequest, RestoreDefaultPolicyResponse, RestoreDefaultReferenceDataRequest,
+    RestoreDefaultReferenceDataResponse, SetPolicyRequest, SetPolicyResponse,
+    SetReferenceDataRequest, SetReferenceDataResponse, TestOpaRequest, TestOpaResponse,
+};
+
+const DEFAULT_MANAGEMENT_SOCK: &str = "127.0.0.1:3001";
+
+impl Tee {
```
sameo, line 25: Why arent we moving those implementations to `crate::management_api::Tee` ?

server/src/users/mod.rs
```diff
@@ -0,0 +1,11 @@
+use std::path::{Path, PathBuf};
+
+pub mod policy;
+
+const DEFAULT_USER_ID: &str = "default";
+
```
sameo, line 6: I think we want to abstract a user through its specific data structure, even though it could be very simple for now:

```Rust
pub struct User {
    id: String,
}
```
 
then add a `Default` and a `from_string` implementations for it.

All the `users::policy` functions would become `User` methods, allowing us to change and expand the `User` definition in the future without having to modify the API and its callers.

server/src/users/policy.rs
```diff
@@ -0,0 +1,79 @@
...
+use tokio::fs;
+
+const POLICY_NAME: &str = "policy.rego";
+const REFERENCE_DATA_NAME: &str = "reference_data.json";
+
+fn workdir(dir: &Path, id: Option<String>, tee: String) -> PathBuf {
+    users::path(dir, id).join(tee)
+}
+
+pub async fn get_policy(dir: &Path, id: Option<String>, tee: String) -> Result<Option<String>> {
```
sameo, line 13: Idiomatic Rust for getters would be `policy()`, not `get_policy`, which makes even more sense if this becomes a `User` method.
Same for `reference_data()` instead of `get_reference_data`.

server/proto/attestation.proto
```diff
@@ -0,0 +1,15 @@
+syntax = "proto3";
+
+package attestation;
+
+message AttestationRequest {
+    bytes evidence = 1;
+    optional string user = 2;
```
sameo, line 7: I would create a `User` gRPC type and use that instead of a raw user ID string. That will allow to isolate our `User` definition from the gRPC API itself.

server/proto/management.proto
```diff
@@ -0,0 +1,72 @@
...
+message TestOpaResponse {
+    bytes status = 1;
+}
+
+service ManagementService {
+    rpc GetPolicy(GetPolicyRequest) returns (GetPolicyResponse) {};
+    rpc GetReferenceData(GetReferenceDataRequest) returns (GetReferenceDataResponse) {};
+    rpc SetPolicy(SetPolicyRequest) returns (SetPolicyResponse) {};
+    rpc SetReferenceData(SetReferenceDataRequest) returns (SetReferenceDataResponse) {};
+    rpc RestoreDefaultPolicy(RestoreDefaultPolicyRequest) returns (RestoreDefaultPolicyResponse) {};
```
sameo, line 68: Could we comment on that part of the API? What is a `Default` policy? Is it specific to a user? Where is it defined?

server/proto/management.proto
```diff
@@ -0,0 +1,72 @@
...
+    rpc RestoreDefaultReferenceData(RestoreDefaultReferenceDataRequest) returns (RestoreDefaultReferenceDataResponse) {};
+    rpc TestOpa(TestOpaRequest) returns (TestOpaResponse) {};
```
sameo, line 70: That one definitely needs a comment here. What is this for? I find it suspicious that we have a test endpoint as part of our main API. Maybe we want to have a separate test service, that could be optionally implemented.

lib/src/core/mod.rs
liangzhou121: Sure, I will add this in a separate PR.

server/src/attestation.rs
liangzhou121: OK, I will update to make the policy/reference_data files' operation interfaces as separate service which will be wrapped inside a RW lock.

server/src/management.rs
sameo: Makes sense, thanks.

server/proto/management.proto
liangzhou121: The default policy is integrated inside the Attestation Service Crate, please refere the following code of sample TEE:
```
    let policy = policy.unwrap_or_else(|| std::include_str!(default_policy!()).to_string());
    let reference_data = reference_data
        .unwrap_or_else(|| std::include_str!(default_reference_data!()).to_string());
```
For sample TEE, it's default Policy is defined in: `lib/src/core/verifier/sample/policy.rego`

It is specific to different type of TEEs, and each user's default Policy/Reference Data are the same.

server/proto/management.proto
liangzhou121: OK, I will remove this testing API.

knrt10:
Hi folks, I am trying to test and run this, is there any steps or doc that I can follow for the same? Thanks

liangzhou121:
> Hi folks, I am trying to test and run this, is there any steps or doc that I can follow for the same? Thanks

Hi, You can follow this PR's `server/README.md` to build/test and familiar it's functionalities.

Cargo.toml
```diff
@@ -1,4 +1,5 @@
 [workspace]
 members = [
   "lib",
+  "server",
 ]
```
knrt10, line 5: Shouldn't there be a blank line?

server/README.md
```diff
@@ -0,0 +1,181 @@
...
+```
+
+Start Attestation Server and specify the listen port of its gRPC service:
+```shell
+./attestation-server --listen 127.0.0.1:1234
+```
+
+If you want to see the runtime log, run:
+```shell
+RUST_LOG=debug ./attestation-server --listen 127.0.0.1:1234
```
knrt10, line 180: There is no `--listen` flag with the current implementation, I think this should be `attestation-sock`.

server/README.md
```diff
@@ -0,0 +1,181 @@
...
+git clone https://github.com/confidential-containers/attestation-service
+cd attestation-service
+cargo build --release
+```
+
+### Run
+
+- For help information, run:
+```shell
+./attestation-server --help
```
knrt10, line 165: Running `cargo build --release` creates binary inside `target/release`. So I guess it should be

`target/release/attestation-server --help`

knrt10 (changes requested):
Hi @liangzhou121. Great work, the server is running fine as expected. Added some comments, PTAL.

sameo (approved):

sameo:
Hey @knrt10 does this look good to you?

knrt10:
@sameo Yep, I think it’s good to merge.

PR #8, opened 2022-06-17
AS: Add the attestation server's client.

jialez0:


server/app/tool/readme.md
```diff
@@ -0,0 +1,119 @@
...
+- set-policy
+- get-reference-data
+- set-reference-data
+- restore-default-policy
+- restore-default-reference-data
+- test-attestation
+
+### get-policy
+
+Get the <TEE> specific evaluation `Policy(.regp)` from Attestation Server."
```
dubek, line 28: 1. note that the `<TEE>` inside regular markdown text is not rendered (it is consider as an raw-HTML, and `<TEE>` is not a valid HTML tag). So either wrap it with backticks, or remove the angle-bracket, like:

    Get the TEE-specific evaluation `Policy(.rego)` from the Attestsion Server.

2. typo: .regp -> .rego
3. remove final double quotes

(these comments apply to several commands below)

server/app/tool/readme.md
```diff
@@ -0,0 +1,119 @@
...
+```SHELL
+./tool set-policy --tee <TEE> --policy <POLICY> [--addr <ADDR>]
+```
+- tee: Specify the target TEE name.
+- policy: The path of local `Policy(.rego)` which will be upload to Attestation Server.
+- addr(optional): Designate the Attestation Server's `management-sock` address. Default is `127.0.0.1:3001`.
+
+### get-reference-data
+
+Get the <TEE> specific `Reference Data(.json)` from Attestation Server.
```
dubek, line 47: is the reference-data printed to stdout? saved to a file? please add explanations.

server/app/tool/readme.md
```diff
@@ -0,0 +1,119 @@
...
+The `tool` supports each TEE's `Policy(.rego)` and `Reference Data(.json)` configuration and testing subcommands, as the following:
+- get-policy
+- set-policy
+- get-reference-data
+- set-reference-data
+- restore-default-policy
+- restore-default-reference-data
+- test-attestation
+
+### get-policy
```
dubek, line 26: is the policy printed to stdout? saved to a file? please add explanations.

server/app/tool/readme.md
```diff
@@ -0,0 +1,119 @@
...
+- tee: Specify the target TEE name.
+- addr(optional): The Attestation Server's `management-sock` address. Default is `127.0.0.1:3001`.
+
+### test-attestation
+
+Test Attestation Server's `attestation` function with the input <EVIDENCE>.
+```SHELL
+./tool test-attestation [--evidence <EVIDENCE>] [--addr <ADDR>]
+```
+- evidence(optional): The evidence which is evaluated by Attestation Server. Default is a imbedded `sample` TEE's evidence.
```
dubek, line 88: 1. typo: a imbedded -> an embedded
2. what does the embedded sample evidence gives us? will anyone set sample policy or sample reference-data and would like to test them?

server/app/tool/readme.md
```diff
@@ -0,0 +1,119 @@
...
+Restore the Attestation Server's <TEE> specific `Reference Data(.json)` to default.
+```SHELL
+./tool restore-default-reference-data --tee <TEE> [--addr <ADDR>]
+```
+- tee: Specify the target TEE name.
+- addr(optional): The Attestation Server's `management-sock` address. Default is `127.0.0.1:3001`.
+
+### test-attestation
+
+Test Attestation Server's `attestation` function with the input <EVIDENCE>.
```
dubek, line 84: The result is printed to stdout? As JSON? Please clarify, maybe even show an example result in the shell code block below.

server/app/tool/src/api.rs
```diff
@@ -0,0 +1,124 @@
...
+        build::COMMIT_HASH,
+        build::BUILD_TIME
+    );
+
+    let matches = App::new("client")
+    .version(version.as_str())
+    .long_version(version.as_str())
+    .author("Confidential-Containers Team")
+    .subcommand(SubCommand::with_name("set-policy")
+        .about("Set the <TEE> specific evaluation `Policy(.regp)` according to the content of <POLICY>.")
```
dubek, line 40: .regp --> .rego (also elsewhere in this files)

server/app/tool/src/api.rs
```diff
@@ -0,0 +1,124 @@
...
+    .get_matches();
+
+    if let Some(matches) = matches.subcommand_matches("get-policy") {
+        management::get_policy_cmd(matches.value_of("tee"), matches.value_of("addr")).await?;
+    }
+
+    if let Some(matches) = matches.subcommand_matches("get-reference-data") {
+        management::get_reference_data_cmd(matches.value_of("tee"), matches.value_of("addr"))
+            .await?;
+    }
```
dubek, line 85: Instead of 6 if statements, can this be arranged as one match block, like [this example of subcommand_name](https://docs.rs/clap/latest/clap/struct.ArgMatches.html#method.subcommand_name)?

server/app/tool/src/attestation.rs
```diff
@@ -0,0 +1,60 @@
+use anyhow::{anyhow, Context, Result};
+use serde_json::json;
+use sha2::{Digest, Sha384};
+use std::fs;
+
+use crate::attestation_api::attestation_service_client::AttestationServiceClient;
+use crate::attestation_api::{AttestationRequest, AttestationResponse};
+
+const ATTESTATION_SOCK: &str = "127.0.0.1:3000";
```
dubek, line 9: `DEFAULT_ATTESTATION_SOCK` would be clearer here.

server/app/tool/src/attestation.rs
```diff
@@ -0,0 +1,60 @@
...
+
+fn evidence() -> Result<String> {
```
dubek, line 11: 1. rename to `sample_evidence()` or something like that. 
2. As I said above, do we need this as part of the application? Maybe it's enough as part of the unit/integration tests?
3. curious: why is this `Result<String>` and not simply `String` ?

knrt10:
Hi @jialez0. Small question, server application supports SEV-SNP, but this tool only supports(as per readme) only tdx and sgx. Is there any specific reason for this?

server/app/tool/src/attestation.rs
```diff
@@ -0,0 +1,60 @@
...
+        "tee": "sample".to_string(),
+        "tee-pubkey": pubkey,
+        "tee-evidence": tee_evidence
+    })
+    .to_string();
+
+    Ok(evidence)
+}
+
+pub async fn attestation_cmd(tee: Option<&str>, socket: Option<&str>) -> Result<()> {
```
dubek, line 43: so `tee` is actually a filename/path? can it be some Rust type that indicates a path?

server/app/tool/src/attestation.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    let request = AttestationRequest {
+        evidence: evidence.into_bytes(),
+        user: None,
+    };
+
+    let mut client = AttestationServiceClient::connect(format!("http://{}", socket)).await?;
+    let response: AttestationResponse = client.attestation(request).await?.into_inner();
+    let results = String::from_utf8(response.attestation_results)?;
+    info!("Attestation Results:\n {}", results);
```
dubek, line 57: Do we need a silent / non-verbose mode that just prints the results without any title/logging stuff?
This will allow piping the result of this command to something else (like jq).

server/app/tool/src/management.rs
```diff
@@ -0,0 +1,159 @@
...
+use std::path::Path;
+
+use crate::management_api::management_service_client::ManagementServiceClient;
+use crate::management_api::{
+    GetPolicyRequest, GetPolicyResponse, GetReferenceDataRequest, GetReferenceDataResponse,
+    RestoreDefaultPolicyRequest, RestoreDefaultReferenceDataRequest, SetPolicyRequest,
+    SetReferenceDataRequest,
+};
+
+const MANAGEMENT_SOCKET: &str = "127.0.0.1:3001";
```
dubek, line 14: Maybe clearer: `DEFAULT_MANAGEMENT_SOCKET`

server/app/tool/src/management.rs
```diff
@@ -0,0 +1,159 @@
...
+    file: Option<&str>,
+    socket: Option<&str>,
+) -> Result<()> {
+    let socket = socket.unwrap_or(MANAGEMENT_SOCKET);
+    if tee.is_none() {
+        return Err(anyhow!("[--tee <TEE>] must be specified."));
+    }
+    if file.is_none() {
+        return Err(anyhow!("[--policy <POLICY>] must be specified."));
+    }
```
dubek, line 47: can this be handled in the command-line parsing part?
and then you don't pass `Option<&str>` to this function.

(same comments for other commands below)

server/app/tool/src/management.rs
```diff
@@ -0,0 +1,159 @@
...
+    let file = dir.join(name);
+    fs::File::create(file.as_path())
+        .context(anyhow!("create file failed"))?
+        .write_all(content.as_bytes())
+        .context(anyhow!("write failed"))
+}
+
+pub async fn set_policy_cmd(
+    tee: Option<&str>,
+    file: Option<&str>,
```
dubek, line 38: can we use a Rust type that indicates a path here, instead of `str`?

(same comments for other commands below)

server/app/tool/src/management.rs
```diff
@@ -0,0 +1,159 @@
...
+    let request = GetPolicyRequest {
+        tee: Tee::from_str(tee)? as i32,
+        user: None,
+    };
+
+    let mut client = ManagementServiceClient::connect(format!("http://{}", socket)).await?;
+    let response: GetPolicyResponse = client.get_policy(request).await?.into_inner();
+    let policy = String::from_utf8(response.content).unwrap();
+    info!("{}", &policy);
+    store(Path::new("./"), "Policy.rego", &policy)?;
```
dubek, line 103: This always saves to `./Policy.rego` ? this is unexpected. Most tools would probably either write to stdout or expect a `--output /path/to/file` argument.

(same comment regarding Reference_data.json below)

dubek (commented):
This looks in the right direction but I have a few comments. See below.

server/app/tool/readme.md
```diff
@@ -0,0 +1,119 @@
...
+Currently, `tool` is a binary package written in Rust. It mainly provides the following functions:
+- Set/Get each TEE's Open Policy Engine(OPA) `Policy(.rego)` and `Reference Data(.json)` files. The `Reference Data(.json)` contents should come from such as Reference Value Provider Service(RVPS), but this item is not reflected in this implementation.
+- Restore each TEE's `Policy(.rego)` and `Reference Data(.json)` to default value.
+- Provide the Attestation Server's `attestation` endpoint testing functionality.
+
+## Supported TEEs
+
+Currently the `tool` supports the following types of TEE:
+- sgx
+- tdx
```
knrt10, line 12: Looking at the code, I think `sev-snp` should also be added here.

server/app/tool/readme.md
jialez0: This sample evidence is just a test case to test whether the base function of attestaion-server work correctly or not, the purpose of setting sample policy and reference data is only to confirm that these two "set" functions can work normally.

jialez0: Sample evidence is a built-in use case for testing the sample verifier of the attestation-service. We consider that users of this tool may only want to test that the basic function of the attestation-service work correctly (which is one of the purposes of this tool). Therefore, this sample evidence is built into the code of the tool to provide users with this function.

server/app/attestation-service-ctl/src/management.rs
```diff
@@ -0,0 +1,150 @@
...
+};
+
+const DEFAULT_MANAGEMENT_SOCKET: &str = "127.0.0.1:3001";
+
+impl Tee {
+    fn from_str(tee: &str) -> Result<Tee> {
+        match tee {
+            "sgx" => Ok(Tee::Sgx),
+            "tdx" => Ok(Tee::Tdx),
+            "sev-snp" => Ok(Tee::SevSnp),
```
knrt10, line 21: The [server sid](https://github.com/confidential-containers/attestation-service/blob/main/server/src/management.rs#L43)e has it named as `sevsnp`, so I guess we should keep it same at both the places. WDYT? Also if updated here, readme would probably need editing

server/app/tool/readme.md
dubek: If the built-in sample evidence is used with a built-in policy and built-in reference data, then we can simply do the same test during unit/integeration tests when building.

I think that the test-attestation command is useful for testing a user-defined policy; my question is -- is the built-in sample evidence useful for testing my own policy?

(please correct me if I'm missing a use case for test-attestation.)

Also -- maybe this command should be called simply `attest` ? (an imperative verb, like `get-policy`)

server/app/attestation-service-ctl/readme.md
```diff
@@ -0,0 +1,129 @@
...
+
+The `attestation-service-ctl` supports each TEE's `Policy(.rego)` and `Reference Data(.json)` configuration and testing subcommands, as the following:
+- get-policy
+- set-policy
+- get-reference-data
+- set-reference-data
+- restore-default-policy
+- restore-default-reference-data
+- test-attestation
+
```
sameo, line 26: I'm going to be a PITA and I apologize in advance, but I think we want to have these commands a bit more organized. Something like that, to be discussed:

```
ctl policy set
ctl policy get
ctl policy restore
ctl reference-data set
ctl reference-data get
ctl reference-data restore
ctl test

```

With that approach we only have 3 top level commands (policy, reference-data and test) and we can add any sub-commands to any of them at any point in time without endlessly expanding our list of commands. I feel it makes for a nicer user experience.

server/app/attestation-service-ctl/readme.md
```diff
@@ -0,0 +1,129 @@
...
+- tee: Specify the target TEE name.
+- addr(optional): The Attestation Server's `management-sock` address. Default is `127.0.0.1:3001`.
+
+### test-attestation
+
+Test Attestation Server's `attestation` function with the input `EVIDENCE` and print the attestation result to stdout.
+```SHELL
+./attestation-service-ctl test-attestation [--evidence <EVIDENCE>] [--addr <ADDR>]
+```
+- evidence(optional): The evidence which is evaluated by Attestation Server. Default is an embedded `sample` TEE's evidence.
```
sameo, line 91: When not using the `sample` evidence, what's the format for `<EVIDENCE>` ?

server/app/attestation-service-ctl/readme.md
```diff
@@ -0,0 +1,129 @@
...
+Restore the Attestation Server's TEE-specific `Reference Data(.json)` to default.
+```SHELL
+./attestation-service-ctl restore-default-reference-data --tee <TEE> [--addr <ADDR>]
+```
+- tee: Specify the target TEE name.
+- addr(optional): The Attestation Server's `management-sock` address. Default is `127.0.0.1:3001`.
+
+### test-attestation
+
+Test Attestation Server's `attestation` function with the input `EVIDENCE` and print the attestation result to stdout.
```
sameo, line 87: Why not simply call that command `attest`? I can see it potentially becoming more than just sending a sample evidence.

server/app/attestation-service-ctl/src/api.rs
```diff
@@ -0,0 +1,159 @@
...
+pub mod common {
+    tonic::include_proto!("common");
+}
+
+#[macro_use]
+extern crate log;
+shadow!(build);
+
+#[tokio::main]
+async fn main() -> Result<()> {
```
sameo, line 24: Shouldnt this file be `main.rs`?

server/app/attestation-service-ctl/src/api.rs
```diff
@@ -0,0 +1,159 @@
...
+    );
+
+    let matches = App::new("client")
+    .version(version.as_str())
+    .long_version(version.as_str())
+    .author("Confidential-Containers Team")
+    .subcommand(SubCommand::with_name("set-policy")
+        .about("Set the <TEE> specific evaluation `Policy(.rego)` according to the content of <POLICY>.")
+        .arg(Arg::from_usage("--addr=[ADDR] 'Attestation Service's `management-sock` address.'"))
+        .arg(Arg::from_usage("--tee=[TEE] 'The target TEE name'"))
```
sameo, line 43: `--addr`  and `--tee`are global parameters that could be processed separately. They're not specific to only a tiny subset of commands.

server/app/attestation-service-ctl/src/api.rs
```diff
@@ -0,0 +1,159 @@
...
+        .init();
+
+    let version = format!(
+        "\nv{}\ncommit: {}\nbuildtime: {}",
+        build::PKG_VERSION,
+        build::COMMIT_HASH,
+        build::BUILD_TIME
+    );
+
+    let matches = App::new("client")
```
sameo, line 36: FWIW, `clap`allows for a very convenient, `derive`-based definitions of commands, subcommands and their related arguments. Here is a good example: https://github.com/clap-rs/clap/blob/v3.2.6/examples/git-derive.rs

Most of the parsing you're doing below would no longer be necessary with that approach.

server/app/attestation-service-ctl/readme.md
jialez0: It depends on which verifier the user wants to test. For example, if the user want to test a TDX verifier, he can prepare a evidence of the format defined by a TDX verifier.

sameo: Right. So should we pass a file path containing that evidence? It'd be easier for complex evidence formats like TDX or SEV.

server/Cargo.toml
```diff
@@ -5,6 +5,10 @@ edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+[[bin]]
+name = "attestation-service-ctl"
+path = "app/attestation-service-ctl/src/api.rs"
```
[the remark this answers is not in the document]
jialez0: I changed it to `tools`, which should be a more appropriate name.

server/tools/attestation-service-ctl/src/attestation.rs
```diff
@@ -0,0 +1,24 @@
...
+    let request = AttestationRequest {
+        evidence: evidence.into_bytes(),
+        user: None,
+    };
+
+    let mut client = AttestationServiceClient::connect(format!("http://{}", socket)).await?;
+    let response: AttestationResponse = client.attestation(request).await?.into_inner();
+    let results = String::from_utf8(response.attestation_results)?;
+    debug!("Attestation Results:\n");
+    info!("{}", results);
```
jodh-intel, line 21: Can you combine these into a single call maybe (multi-line log output is confusing).

server/tools/attestation-service-ctl/readme.md
```diff
@@ -0,0 +1,34 @@
...
+- Restore each TEE's `Policy(.rego)` and `Reference Data(.json)` to default value.
+- Provide the Attestation Server's `attestation` endpoint testing functionality.
+
+## Supported TEEs
+
+Currently the `attestation-service-ctl` supports the following types of TEE:
+- sgx
+- tdx
+- sevsnp
+- sample: The TEE used to demo/test Attestation Server's functionalities.
```
jodh-intel, line 14: What does this actually mean though? Is it a "dummy" / nop TEE? Maybe the name should make it's functionality (or lack of?) clearer to avoid confusion.

server/tools/attestation-service-ctl/readme.md
```diff
@@ -0,0 +1,34 @@
+# attestation-service-ctl
```
jodh-intel, line 1: Nit: It's more usual for this filename to be upper-case: `README.md`.

server/tools/attestation-service-ctl/readme.md
```diff
@@ -0,0 +1,34 @@
...
+### Build
+
+Build the attestation-service-ctl and Attestation Server.
+```shell
+git clone https://github.com/confidential-containers/attestation-service
+cd attestation-service
+cargo build --release
+```
+
+### Run
```
jodh-intel, line 29: Aside from how to run it, it would be helpful if this document also explained:

- _Who_ should run it.
-  Minimum privilege level it can be run with.
- _Where_ it should be run (can it connect remotely to another system?)

server/tools/attestation-service-ctl/src/attestation.rs
```diff
@@ -0,0 +1,24 @@
+use anyhow::{anyhow, Result};
+use std::fs;
+use std::path::Path;
+
+use crate::attestation_api::attestation_service_client::AttestationServiceClient;
+use crate::attestation_api::{AttestationRequest, AttestationResponse};
+
+pub async fn attestation_cmd(evidence_path: &Path, socket: &str) -> Result<()> {
```
jodh-intel, line 8: The value passed to this function isn't a socket, it's an IPv4 address fwics, so this parameter name is confusing.

server/tools/attestation-service-ctl/src/attestation.rs
```diff
@@ -0,0 +1,24 @@
...
+    let evidence =
+        fs::read_to_string(evidence_path).map_err(|e| anyhow!("Read evidence error: {:?}", e))?;
+
+    let request = AttestationRequest {
+        evidence: evidence.into_bytes(),
+        user: None,
+    };
+
+    let mut client = AttestationServiceClient::connect(format!("http://{}", socket)).await?;
```
jodh-intel, line 17: What if the specified address is a non-local one? Does this have to be "http" or can it be "https"?

jodh-intel (commented):
Thanks @jialez0 - a few comments...

sameo (approved):
Thanks a lot for addressing my comments @jialez0. Besides @jodh-intel , this LGTM now.

server/tools/attestation-service-ctl/src/attestation.rs
jialez0: This can connect to Attestation-Server deployed locally or remotely, and both "http" and "https" is supported.

dubek: So maybe the default should be `"http://127.0.0.1:3001"` and the user should specify the protocol (URL scheme) as well when using the `--addr http://my-service.com:12345` option.
(and then we don't need the `format!()` call here)

server/tools/attestation-service-ctl/src/management.rs
```diff
@@ -0,0 +1,114 @@
...
+pub async fn get_policy_cmd(tee: &str, output_path: &Path, socket: &str) -> Result<()> {
+    let request = GetPolicyRequest {
+        tee: Tee::from_str(tee)? as i32,
+        user: None,
+    };
+
+    let mut client = ManagementServiceClient::connect(format!("http://{}", socket)).await?;
+    let response: GetPolicyResponse = client.get_policy(request).await?.into_inner();
+    let policy = String::from_utf8(response.content).unwrap();
+    info!("{}", &policy);
```
fitzthum, line 65: Do we want to be using the logger here rather than just printing to stdout?

server/tools/attestation-service-ctl/src/management.rs
```diff
@@ -0,0 +1,114 @@
...
+pub async fn set_policy_cmd(tee: &str, file: &Path, socket: &str) -> Result<()> {
+    let policy = fs::read_to_string(file).context(anyhow!("Read policy error"))?;
+
+    let request = SetPolicyRequest {
+        tee: Tee::from_str(tee)? as i32,
+        user: None,
+        content: policy.into_bytes(),
+    };
+
+    let mut client = ManagementServiceClient::connect(format!("http://{}", socket)).await?;
```
fitzthum, line 37: Ideally we wouldn't have to do this same format every single time. Also, the variable `socket` here is more like an address than a socket.

fitzthum (approved):
A few minor points but looks good.

sameo:
@jodh-intel Does that look good to you now?

jodh-intel (approved):
Thanks @jialez0.

PR #12, opened 2022-07-20
Refactoring the integration logic of the policy engine.

jialez0:
The policy engine used in the Attestation Service is the powerful and flexible [Open Policy Agent (OPA)](https://www.openpolicyagent.org/), it can provide a flexible, fine-grained control for the owner of Attestation Service. 

This PR optimizes the integration logic of OPA and Attestation Service, and turns the policy engine into a general functional component on the upper layer of the verifier driver, which is used to verify whether the TCB status of various HW-TEE meets the owner's expectations.


lib/src/lib.rs
```diff
@@ -278,25 +244,24 @@ allow {
...
+        println!("{:?}", res.as_ref());
         assert!(res.is_ok(), "attestation should success");
         let v: Value = serde_json::from_str(&res.unwrap()).unwrap();
-        assert!(v["result"] == "true".to_string(), "result should true");
-        let policy_info: Value = serde_json::from_str(v["policy_info"].as_str().unwrap()).unwrap();
-        assert!(
-            policy_info["policy_name"] == "policy.rego".to_string(),
-            "policy name should be policy.rego"
-        );
+        assert!(v["result"] == json!(true), "result should true");
```
dubek, line 256: if `assert!` prints the expression (upon error), then there's no need for the "reason" string (2nd argument), because it doesn't explain a lot more than the expression itself) (and same for line 264 below).

server/README.md
```diff
@@ -57,35 +57,29 @@ enum Tee {
...
 #### GetPolicy
 
-Get the specific TEE's OPA `Policy(.rego)`. The message:
+Get the OPA `Policy(.rego)`. The message:
 ```PROTO
 message GetPolicyRequest {
-    // Supported TEE types
-    Tee tee = 1;
     // Optional: Designate the user id. It should be kept as "None" currently.
     optional string user = 2;
```
dubek, line 64: This fields remains `= 2` to avoid confusing client that already used `1` for the `tee` field?
(same question regarding all messages here)

dubek:
Question on this PR: If I have a deployment with containers running both in SNP and in TDX, do I need to run two separate attestation-servers, and configure each KBC to connect to "its own kind" attestation-server? Because I will surely need a separate reference data and policy for each TEE.

jialez0:
> Question on this PR: If I have a deployment with containers running both in SNP and in TDX, do I need to run two separate attestation-servers, and configure each KBC to connect to "its own kind" attestation-server? Because I will surely need a separate reference data and policy for each TEE.

Not at all. This is exactly the meaning of this PR. OPA is a very powerful engine, far more powerful than our previous usage, so it can become a general policy verification function module in AS. It completely allows us to use a universal policy.rego file to cover the verification of different input data of all TEE. In this mode, the verify drivers of different TEE only need to verify the hardware signature in the evidence, Then analyze those fields that need the policy engine to perform the next verification and return them in a JSON claim. The general policy engine can verify the JSON claims returned by different TEE verify drivers. The default policy.rego file provided in the PR can do this. Users can also write more complex and powerful policy.rego files to achieve some very delicate verification.

dubek:
>  The general policy engine can verify the JSON claims returned by different TEE verifiers. The default policy.rego file provided in this PR can do this. Users can also write more complex and powerful policy.rego files to achieve some very delicate verification.

So my reference data should include something like:

```
{
  "snp_measurement": ["ABCDEF", "123456"],
  "tdx_measurement": ["FEDC", "9876"]
}
```

and then if the quote produce by the SNP verifier has `{"snp_measurement": "123456"}` it will match the default policy rules and will be allowed. Whereas the TDX verifier will produce "tdx_measurement" which will have different possible values (or produce several fields, etc.).

Did I understand this correctly?

jialez0:
> > The general policy engine can verify the JSON claims returned by different TEE verifiers. The default policy.rego file provided in this PR can do this. Users can also write more complex and powerful policy.rego files to achieve some very delicate verification.
> 
> So my reference data should include something like:
> 
> ```
> {
>   "snp_measurement": ["ABCDEF", "123456"],
>   "tdx_measurement": ["FEDC", "9876"]
> }
> ```
> 
> and then if the quote produce by the SNP verifier has `{"snp_measurement": "123456"}` it will match the default policy rules and will be allowed. Whereas the TDX verifier will produce "tdx_measurement" which will have different possible values (or produce several fields, etc.).
> 
> Did I understand this correctly?

That's what I mean. Field names can be more refined, for example: `snp-cpu-svn`, `tdx-mrsigner` and so on.

jialez0:
> > The general policy engine can verify the JSON claims returned by different TEE verifiers. The default policy.rego file provided in this PR can do this. Users can also write more complex and powerful policy.rego files to achieve some very delicate verification.
> 
> So my reference data should include something like:
> 
> ```
> {
>   "snp_measurement": ["ABCDEF", "123456"],
>   "tdx_measurement": ["FEDC", "9876"]
> }
> ```
> 
> and then if the quote produce by the SNP verifier has `{"snp_measurement": "123456"}` it will match the default policy rules and will be allowed. Whereas the TDX verifier will produce "tdx_measurement" which will have different possible values (or produce several fields, etc.).
> 
> Did I understand this correctly?

That's what I mean. Field names can be more refined, for example: `snp-cpu-svn`, `tdx-mrsigner` and so on.

.github/workflows/basic.yml
```diff
@@ -17,6 +17,16 @@ jobs:
...
           fetch-depth: 1
 
+      - name: Instll OPA command line tool
+        run: |
+          curl -L -o opa https://openpolicyagent.org/downloads/v0.42.2/opa_linux_amd64_static
+          chmod 755 ./opa && cp opa /usr/local/bin
+      
+      - name: OPA policy.rego fmt and check
+        run: |
+          opa fmt ./lib/src/core/policy_engine/default_policy.rego
```
dubek, line 27: I think this should be:

```
opa fmt --list --fail ...
```

(but I haven't tried it)

dubek (approved):
Looks good, see small comment below:

lib/src/core/policy_engine/default_policy.rego
```diff
@@ -0,0 +1,43 @@
...
+	input_value == reference_value
+}
+
+match_value(reference_value, input_value) {
+	is_array(reference_value)
+	array_include(reference_value, input_value)
+}
+
+array_include(reference_value_array, input_value) {
+	reference_value_array == []
```
fitzthum, line 34: So any claim will be true if it is made against a reference array that is empty? This is what we do in `simple-kbs` as well (because it's hard to define a good wildcard value). Let's just make sure we document this clearly because it could lead to accidental policy validation.

lib/src/core/mod.rs
```diff
@@ -66,17 +65,56 @@ impl Attestation {
...
         let verifier = TEE::from_str(&evidence.tee)?.to_verifier()?;
-        let results = verifier.evaluate(&evidence, policy, reference_data).await?;
-        let results = serde_json::to_string(&results)?;
-        debug!("Attestation Results: {}", &results);
-        Ok(results)
-    }
 
-    pub fn policy(&self, tee: TEE) -> Result<String> {
-        tee.to_verifier()?.default_policy()
+        let tee_evidence_parsed_claims = match verifier.evaluate(&evidence).await {
```
fitzthum, line 69: might be more clear as `claims_parsed_from_tee_evidence` or just `claims_from_evidence`. Maybe also change the type alias above.

fitzthum (approved):
Great work! This simplifies a lot of our interfaces and should make integration with the RVPS much simpler. 

There is also some nice general cleanup in this PR. 

I made a few suggestions but looks good overall.

lib/src/core/policy_engine/default_policy.rego
jialez0: Yes, this default policy needs to be written into documents to explain to users, but now is not the time because AS is still in the early construction stage. Let's remember it and explain this when adding documents to AS.

sameo (approved):

Issue #11, opened 2022-07-18
Reference Value Provider Service Crate


Xynnn007:
As the proposal https://github.com/confidential-containers/documentation/issues/37,
RVPS is a key component for AS to gather the `trusted digests`.

ariel-adam:
@Xynnn007 is this issue still relevant or can be closed? 
If it's still relevant to what release do you think we should map it to (mid-November, end-December, mid-February etc...)?

Xynnn007:
@ariel-adam I think this issue can not be closed until all the tasks are finished. If RVPS is finally integrated into AS, I will close this issue, thx

PR #13, opened 2022-07-22
feat: Implement RVPS (Reference Value Provider Service)

Xynnn007:
Proposal refer to: https://github.com/confidential-containers/documentation/issues/37
Issue refer to: https://github.com/confidential-containers/attestation-service/issues/11

This pr includes
Basic architecture of the RVPS, including

- Pre-Processor
- Extractors
- Cache

Also define the following formats:

- Provenance Message
- Reference Value


jialez0 (approved):
LGTM, thanks a lot!

rvps/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+/// * `version`: version of this message.
+/// * `payload`: content of the provenance, JSON encoded.
+/// * `typ`: provenance type of the payload.
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Message {
+    #[serde(default = "default_version")]
+    version: String,
+    payload: String,
+    #[serde(rename = "type")]
+    typ: String,
```
jodh-intel, line 38: If you changed the field to `r#type`, I think you can remove the serde annotation.

rvps/src/pre_processor/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+pub struct PreProcessor {
+    wares: Vec<Box<dyn Ware>>,
+}
+
+impl PreProcessor {
+    pub fn new() -> Self {
+        Self { wares: Vec::new() }
+    }
```
jodh-intel, line 71: Derivable?

rvps/src/reference_value.rs
```diff
@@ -0,0 +1,185 @@
...
+
+/// Helper to deserialize an expired time
+fn primitive_date_time_from_str<'de, D: Deserializer<'de>>(
+    d: D,
+) -> Result<DateTime<Utc>, D::Error> {
+    let s: Option<String> = Deserialize::deserialize(d)?;
+    if s.is_none() {
+        return Err(serde::de::Error::invalid_length(0, &"<TIME>"));
+    }
+    let s = s.unwrap();
```
jodh-intel, line 45: This would be cleaner using `map_err(...)?` or similar.

jodh-intel (commented):
Thanks @Xynnn007 - a few comments...

rvps/src/extractors/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+/// `Extractors` provides different kinds of `Extractor`s due to
+/// different provenance types, e.g. in-toto, etc.
+/// Each `Extractor` will process the input provenance, verify the
+/// authenticity of the provenance, and then extract the formatted
+/// reference value (degest, s.t. hash value and name of the artifact)
+/// from the provenance. If the verification fails, no reference value
+/// will be extracted.
+
+/// `ExtractorsAPI` defines the interfaces of Extractors.
+pub trait ExtractorsAPI {
```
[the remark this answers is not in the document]
Xynnn007: Surely!

rvps/src/extractors/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+}
+
+impl Extractors {
+    pub fn new() -> Self {
+        let extractors_module_list = ExtractorModuleList::new();
+        let extractors_instance_map = HashMap::new();
+        Extractors {
+            extractors_module_list,
+            extractors_instance_map,
+        }
```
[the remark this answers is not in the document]
Xynnn007: Yes, all can be default derived. The following as the same

rvps/src/lib.rs
Xynnn007: Yea, both are OK. I get accustomed to extra serde annotation because I somehow do not like `r#` here  :P.  I'll change it as your suggestion

rvps/src/reference_value.rs
Xynnn007: Sure!

rvps/src/pre_processor/mod.rs
Xynnn007: Yes, as default

Xynnn007:
@jodh-intel Thanks for your suggestions! Have changed as them

jialez0:
@jodh-intel Does this look good to you now?

rvps/src/reference_value.rs
```diff
@@ -0,0 +1,182 @@
...
+/// This Reference Value is not the same as Reference in IETF's RATS.
+/// Here, RV is consumed by AS. Its format MAY be modified often to
+/// cowork with AS.
+/// * `version`: version of the reference value format.
+/// * `name`: name of the artifact related to this reference value.
+/// * `expired`: expired time for this reference value.
+/// * `hash_value`: A set of key-value pairs, each indicates a hash
+/// algorithm and its relative hash value for the artifact.
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
+pub struct ReferenceValue {
```
jialez0, line 60: A little suggestion: I'm not sure if we really need such a heavy RV structure, any fields here may be unnecessary, such as `version`, `expired` and the `algo` in `HashValuePair`. 

For `version`, since this structure is not transmitted in cross process communication, it only needs to be consistent with the AS version. It is unnecessary to list it separately here.

For `expired`, I guess there will be a so-called "subscription" mechanism in the AS to update the reference value from the upstream in time. For the caller, the latest reference value will be obtained from the rvps every time the comparison verification is performed. Therefore, the 'expired' here may not be used.

For `algo`, because the reference value provided by rvps is mainly used for reference, and in fact we do not know what algorithm is used for the compared real value. In comparison and verification, we only need to ensure that they are the same. Therefore, even if there are reference values of multiple algorithms, we only need to simply compare whether the real value is equal to one of them. Therefore, it is not necessary to involve what algorithm they use.

So, perhaps a simple RV structural design can meet all requirements:

```rust
ReferenceValue {
    // The resource name of this RV.
    name: String,
    // The reference hash values, base64 coded.
    hash_values: Vec<String>,
}
```

I'm not sure if this is the best, if I missed something, please remind me :)

Xynnn007: Thanks for your deep dive into the code!

Certainly we can delete these fields, however:
* `version` may give some extensibility for possible changes in the future
* Although `expired` can give an expired time because provenance information like `in-toto` will declare "when this metadata will be expired" without using the publishing mechanism you've mentioned, I think it is more suitable to handle inside RVPS.
*  `algo` is the most controversial field, I preferred to record the digest algorithm for potential usage and extensibility. Maybe it's wiser to delete this for current implementation. In the future, perhaps, we can design a new format to deal with concrete problems by change `version` field.

Overall, I'm with your design for the RV transferred between AS and RVPS.

jialez0: Yes, this information is indeed what the RVPS should record and maintain. My suggestions here are only for the structure transmitted by the RVPS to the AS :)

jialez0:
@Xynnn007 Thanks a lot!

rvps/src/cache/mod.rs
```diff
@@ -0,0 +1,23 @@
...
+use anyhow::Result;
+
+/// Interface of an Cache.
+/// We only provide a simple instance here which implements
+/// Cache. In more scenerios, RV should be stored in persistent
+/// storage, like database, file and so on. All of the mentioned
+/// forms will have the same interface as following.
+pub trait Cache {
+    /// Store a reference value
+    fn set(&mut self, name: String, rv: ReferenceValue) -> Result<()>;
```
jodh-intel, line 19: Would this return `Err` if the value already existed in the cache? If not, shouldn't this return the existing value? That might be useful to the caller, or not (they could discard it if not). Plus, returning the original value would help with uniting testing the code.

rvps/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+use chrono::{DateTime, Utc};
+use extractors::{Extractors, ExtractorsImpl};
+use pre_processor::{PreProcessor, PreProcessorAPI, Ware};
+use serde::{Deserialize, Serialize};
+
+pub use cache::Cache;
+pub use reference_value::{ReferenceValue, TrustedDigest};
+
+/// Default version of Message
+static MESSAGE_VERSION: &str = "0.1";
```
jodh-intel, line 25: - What will cause this value to change?
- Why not use [semver](https://semver.org) for the version?

rvps/src/reference_value.rs
```diff
@@ -0,0 +1,197 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+//! reference value for RVPS
+
+use chrono::{DateTime, NaiveDateTime, Utc};
+use serde::{Deserialize, Deserializer, Serialize};
+
+/// Default version of ReferenceValue
+pub const REFERENCE_VALUE_VERSION: &str = "0.1";
```
jodh-intel, line 12: Same comment as for `MESSAGE_VERSION`: why not use [semver](https://semver.org) here?

rvps/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+        self.pre_processor.add_ware(ware);
+        self
+    }
+}
+
+impl<T: Cache> RVPSAPI for Core<T> {
+    fn verify_and_extract(&mut self, mut message: Message) -> Result<()> {
+        // Judge the version field
+        if message.version != MESSAGE_VERSION {
+            return Err(anyhow!(
```
jodh-intel, line 89: Could you add a unit test to assert this error occurs?

jodh-intel (commented):
Thanks @Xynnn007 - a few comments...

rvps/src/lib.rs
Xynnn007: As proposed, a `Message` here is used to convey different kinds of provenances.  It is exactly `Provenances` in [this picture](https://user-images.githubusercontent.com/22250460/173334824-678f32cf-dcbe-437f-b44d-bbe439b79a29.png) in [the proposal](https://github.com/confidential-containers/documentation/issues/37). Because the `Message` here is somehow a conceptual message type, I think it is rational to keep a field as `version` for potential extensibility demand.

rvps/src/lib.rs
Xynnn007: Thanks to your suggestion, I think this error occurs when the writer edited the `version` field by mistake. If a different layout message is sent, certainly it can not be deserialized, thus not able to reach here.

I will add the unit test in the next pr. Because I didn't add any implementation for `Cache`, which is needed to create a RVPS's `Core` object, the test is now not able to be written.

Xynnn007:
@jodh-intel Thanks for your great thinking and suggestions. Please take some time to review it again...

jodh-intel (approved):
Thanks @Xynnn007.

rvps/src/extractors/extractor_modules/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+use anyhow::*;
+use std::collections::HashMap;
+
+use crate::reference_value::ReferenceValue;
+
+/// Extractor is a standard interface that all provenance extractors
+/// need to implement. Here reference_value can be modified in the
+/// handler, added any field if needed.
+pub trait Extractor {
+    fn verify_and_extract(&self, provenance: &str) -> Result<ReferenceValue>;
```
fitzthum, line 17: So one piece of provenance will always result in just one reference value? There's no way to have some kind of manifest that accounts for many different components?

fitzthum (commented):
Looks good. A few comments. Some of this might become more clear in future PRs.

rvps/src/cache/mod.rs
```diff
@@ -0,0 +1,23 @@
...
+use anyhow::Result;
+
+/// Interface of an Cache.
+/// We only provide a simple instance here which implements
+/// Cache. In more scenerios, RV should be stored in persistent
+/// storage, like database, file and so on. All of the mentioned
+/// forms will have the same interface as following.
+pub trait Cache {
+    /// Store a reference value
+    fn set(&mut self, name: String, rv: ReferenceValue) -> Result<Option<ReferenceValue>>;
```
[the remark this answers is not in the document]
Xynnn007: Inspired by @jodh-intel , if we `set` new `rv` with an existed `name` , the caller can decide how to deal with the old `rv`, which means either to abandon directly or do some other things.

rvps/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+/// message, for extendability).
+/// * `version`: version of this message.
+/// * `payload`: content of the provenance, JSON encoded.
+/// * `type`: provenance type of the payload.
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Message {
+    #[serde(default = "default_version")]
+    version: String,
+    payload: String,
+    r#type: String,
```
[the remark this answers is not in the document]
Xynnn007: For `type` is a key word in Rust. If we use another name like `typ`, there should be a `#[serde(rename = "type")]` macro. IMO, it might not be a big problem, for different habits for persons.

rvps/src/pre_processor/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+            Ok(())
+        }
+    }
+}
+
+/// PreProcessor's interfaces
+/// `process` processes a `provenance` of type `typ`,
+/// and will output a tuple (provenance, typ)
+pub trait PreProcessorAPI {
+    fn process(&self, message: &mut Message) -> Result<()>;
```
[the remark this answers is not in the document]
Xynnn007: Thanks for your deep dive here! There is a trade-off I considered for a long time. 

I wanted the `process` interface here to do some processing work for a `Message`, which will not consume the `Message` object. There might be two design for the interface:
* `fn process(&self, message: &mut Message) -> Result<()>;`
* `fn process(&self, message: Message) -> Result<Message>;`

Both of them receives a `Message`, and when `process()` is done, the `Message` will be processed.
I prefer the first one, although it looks some C-style. I'd very like to know your opinions upon the interface design and selection

rvps/src/extractors/extractor_modules/mod.rs
Xynnn007: Wow, you are right. Let me update this as `Result<Vec<ReferenceValue>>`

Xynnn007:
@fitzthum Thanks for your deep dive and suggestions! There are three conversations left, please take some time to share your opinion

sameo (approved):

jialez0:
@fitzthum Hi, does this PR look good to you now?

rvps/src/pre_processor/mod.rs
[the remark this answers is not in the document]
fitzthum: Ok, if it's just modifying the message itself I think the first option is fine. The comment makes it seem like the preprocessor is going to output a tuple, but I guess it is transforming the messages.

fitzthum (approved):
LGTM.

rvps/src/pre_processor/mod.rs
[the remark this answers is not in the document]
Xynnn007: Thanks, sry for the bug about docs. I'll fix it

PR #15, opened 2022-08-12
Allow "clippy::derive_partial_eq_without_eq" for autogenerated code

jialez0:
This is a recommendation given by the maintainer of the prost crate.
See: tokio-rs/prost#661

The additional `#![allow(unknown_lints)]` had to be added as the
`derive_partial_eq_without_eq` is unknown by the `stable` and `beta`
toolchains.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+target/
+Cargo.lock
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -10,12 +10,18 @@ use tokio::sync::RwLock;
 extern crate lazy_static;
 
 pub mod management_api {
+    #![allow(unknown_lints)]
+    #![allow(clippy::derive_partial_eq_without_eq)]
     tonic::include_proto!("management");
 }
 pub mod attestation_api {
+    #![allow(unknown_lints)]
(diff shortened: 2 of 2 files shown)


jialez0:
Does anyone know why the `#![allow(clippy::derive_partial_eq_without_eq)]` can not resolve the clippy check issue now?

fidencio:
@stevenhorsman, tagging you here as you just solved one of this recently.

stevenhorsman:
Sorry for the delayed comment - I was having fun trying a bunch of attempted fixes. So I didn't actually manage to fix this earlier, but came to the conclusion (which could well be wrong) that as the problem was in the tonic generated code then it required the fix (either to derive Eq, or add the `#![allow(clippy::derive_partial_eq_without_eq)]`) to be done in tonic. Based on the comment from the maintainer that they'd recommend ignoring the warning for now I assume that won't happen very soon. The only way I know to get around this problem is to disable it completely from clippy, which seems to work from me in https://github.com/confidential-containers/attestation-service/pull/16. Not ideal, but I don't currently know of another option. I'm sure smarter people than me might be able to come up with something better though.

jialez0:
@stevenhorsman OK, at present, #16 really seems to be a good solution. Thanks!

PR #17, opened 2022-09-20
[RVPS] Implement a local file storage as Cache

Xynnn007:
Due to https://github.com/confidential-containers/attestation-service/issues/11, an implementation of local file storage as `Store` is added in this pr. 

#### Change Log

- Added local-fs storage using `sled` as engine.
- Added related document and interface tests of local-fs storage.
- All the data will be stored in the directory `/opt/attestation-server/reference_values`, where `/opt/attestation-server` is const `ATTESTATION_SERVER_WORKDIR ` in attestation-server.

### DIFF
--- a/rvps/src/store/local_fs/mod.rs
+++ b/rvps/src/store/local_fs/mod.rs
@@ -1,0 +1,158 @@
+// Copyright (c) 2022 Alibaba Cloud
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+//! This Store stores RV information inside a local file
+
+use std::path::Path;
+
+use anyhow::*;
+
+use crate::{ReferenceValue, Store};
+
+/// Local directory path to store the reference values,
+/// which is created by sled engine.
+const FILE_PATH: &str = "/opt/attestation-server/reference_values";
+
+/// `LocalFs` implements [`Store`] trait. And
+/// it uses rocksdb inside.
+pub struct LocalFs {
+    engine: sled::Db,
+}
+
+impl Default for LocalFs {
+    /// Create a `LocalFs` storage, which will use path [`FILE_PATH`]
+    /// to store files.
+    fn default() -> Self {
+        let path = Path::new(FILE_PATH);
+        LocalFs::new(path).expect("Failed to create LocalFs Store.")
+    }
+}
+
+impl LocalFs {
+    /// Create a new [`LocalFs`] with given
+    /// file storage path.
+    fn new(path: &Path) -> Result<Self> {
+        let engine = sled::open(path)?;
+        Ok(Self { engine })
+    }
+}
+
+impl Store for LocalFs {
+    fn set(&mut self, name: String, rv: ReferenceValue) -> Result<Option<ReferenceValue>> {
+        let rv_serde = serde_json::to_vec(&rv)?;
+        let res = match self.engine.insert(name, rv_serde)? {
+            Some(v) => {
+                let v = serde_json::from_slice(&v)?;
+                Ok(Some(v))
+            }
+            None => Ok(None),
+        };
+
+        self.engine.flush()?;
+        res
+    }
+
+    fn get(&self, name: &str) -> Result<Option<ReferenceValue>> {
+        match self.engine.get(name)? {
+            Some(v) => {
+                let v = serde_json::from_slice(&v)?;
+                Ok(Some(v))
+            }
+            None => Ok(None),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use serial_test::serial;
+
+    use crate::{store::local_fs::LocalFs, ReferenceValue, Store};
+
+    const KEY: &str = "test1";
+
+    /// This test will test the `set` and `get` interface
+    /// for [`LocalFs`].
+    #[test]
+    #[serial]
+    fn set_and_get() {
+        let temp_dir = tempfile::tempdir().expect("create tempdir failed");
+        {
+            let mut store = LocalFs::new(temp_dir.path()).expect("create local fs store failed.");
+            let rv = ReferenceValue::new().expect("create ReferenceValue failed.");
+            assert!(
+                store
+                    .set(KEY.to_owned(), rv.clone())
+                    .expect("set rv failed.")
+                    .is_none(),
+                "the storage has previous key of {}",
+                KEY
+            );
+            let got = store
+                .get(KEY)
+                .expect("get rv failed.")
+                .expect("get None from LocalFs Store");
+            assert_eq!(got, rv);
+        }
+    }
+
+    /// This test will test the `set` interface with the
+    /// duplicated key for [`LocalFs`].
+    #[test]
+    #[serial]
+    fn set_duplicated() {
+        let temp_dir = tempfile::tempdir().expect("create tempdir failed");
+        {
+            let mut store = LocalFs::new(temp_dir.path()).expect("create local fs store failed.");
+            let rv_old = ReferenceValue::new()
+                .expect("create ReferenceValue failed.")
+                .set_name("old");
+
+            let rv_new = ReferenceValue::new()
+                .expect("create ReferenceValue failed.")
+                .set_name("new");
+
+            assert!(
+                store
+                    .set(KEY.to_owned(), rv_old.clone())
+                    .expect("set rv failed.")
+                    .is_none(),
+                "the storage has previous key of {}",
+                KEY
+            );
+
+            let got = store
+                .set(KEY.to_owned(), rv_new)
+                .expect("get rv failed.")
+                .expect("get None from LocalFs Store");
+
+            assert_eq!(got, rv_old);
+        }
+    }
+
+    /// This test will simulate a restart operation
+    /// for [`LocalFs`].
+    #[test]
+    #[serial]
+    fn restart() {
+        let rv = ReferenceValue::new().expect("create ReferenceValue failed.");
+        let temp_dir = tempfile::tempdir().expect("create tempdir failed");
+        {
+            let mut store = LocalFs::new(temp_dir.path()).expect("create local fs store failed.");
+            store
+                .set(KEY.to_owned(), rv.clone())
+                .expect("set rv failed.");
+        }
+        {
+            let store =
+                LocalFs::new(temp_dir.path()).expect("read previous local fs store failed.");
+            let got = store
+                .get(KEY)
+                .expect("get rv failed.")
+                .expect("get None from LocalFs Store");
+            assert_eq!(got, rv);
+        }
+    }
+}
--- a/rvps/src/store/mod.rs
+++ b/rvps/src/store/mod.rs
@@ -9,13 +9,16 @@ use crate::reference_value::ReferenceValue;
 
 use anyhow::Result;
 
+pub mod local_fs;
+
 /// Interface of a `Store`.
 /// We only provide a simple instance here which implements
 /// Store. In more scenarios, RV should be stored in persistent
 /// storage, like database, file and so on. All of the mentioned
 /// forms will have the same interface as following.
 pub trait Store {
-    /// Store a reference value
+    /// Store a reference value. If the given `name` exists,
+    /// return the previous `Some<ReferenceValue>`, otherwise return `None`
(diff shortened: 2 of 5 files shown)


rvps/src/cache/local_fs/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+        let path = Path::new(path);
+        let engine = sled::open(path)?;
+        Ok(Self { engine })
+    }
+}
+
+impl Cache for LocalFs {
+    fn set(&mut self, name: String, rv: ReferenceValue) -> Result<Option<ReferenceValue>> {
+        let rv_serde = serde_json::to_string(&rv)?;
+        println!("{}", rv_serde);
```
dubek, line 43: I assume this is debugging leftover?

rvps/src/cache/local_fs/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+    fn new(path: &str) -> Result<Self> {
+        let path = Path::new(path);
+        let engine = sled::open(path)?;
+        Ok(Self { engine })
+    }
+}
+
+impl Cache for LocalFs {
+    fn set(&mut self, name: String, rv: ReferenceValue) -> Result<Option<ReferenceValue>> {
+        let rv_serde = serde_json::to_string(&rv)?;
```
dubek, line 42: I think we can use `serde_json::to_vec()` here and then no need for `rv_serde.to_bytes()` below.

rvps/src/cache/local_fs/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+        println!("{}", rv_serde);
+        match self.engine.insert(name, rv_serde.as_bytes())? {
+            Some(v) => {
+                self.engine.flush()?;
+                let v = serde_json::from_slice(&v)?;
+                Ok(Some(v))
+            }
+            None => Ok(None),
+        }
```
dubek, line 51: Here we flush if there a previous value already existed in the DB, but we don't flush if it's a new key-value pair. That doesn't seem like the correct logic.

Also, is there a use to returning the old value? Is that the expected behaviour of `Cache::set()` ?
If there's no use, maybe we can return `Result<()>` (here and in `cache/mod.rs`).

rvps/src/cache/local_fs/mod.rs
Xynnn007: Yes, sry for that. Fixed it, thanks

rvps/src/cache/local_fs/mod.rs
Xynnn007: Yes, thanks

rvps/src/cache/local_fs/mod.rs
Xynnn007: Yes, it is the expected behavior of `Cache::set()`. I've add some more notes on the trait.

dubek: > > Also, is there a use to returning the old value? Is that the expected behaviour of `Cache::set()` ?
> > If there's no use, maybe we can return `Result<()>` (here and in `cache/mod.rs`).
> 
> Yes, it is the expected behavior of `Cache::set()`. I've add some more notes on the trait.

If that's the expected behaviour, it should be tested in the unit-tests. It seems that in `set_and_get` we ignore the return value of `.set()`. Consider adding a test for setting a new key and for setting an existing key, and checking the return value in both cases.

Xynnn007: Yes, you're right. I've added it, thanks!

rvps/src/reference_value.rs
```diff
@@ -75,13 +76,19 @@ fn default_version() -> String {
 }
 
 impl ReferenceValue {
-    pub fn new() -> Self {
-        ReferenceValue {
+    /// Create a new `ReferenceValue`, the `expired`
+    /// field's nanosecond will be set to 0. This avoid
+    /// a rare bug that when the nanosecond of the time
+    /// is not 0, the test case will fail.
```
dubek, line 82: I think the bug can also hit us if user calls `.set_expired(Utc.ymd(2022, 9, 22).and_hms(8, 59, 59).with_nanoseconds(123)))` -- in such case serializing and deserializing will not return the original RV.  So maybe `set_expired` should be:

```rust
    pub fn set_expired(mut self, expired: DateTime<Utc>) -> Self {
        self.expired = expired.with_nanoseconds(0);
        self
    }
```

This way we make sure that the internal `expired` field of `ReferenceValue` always has 0 nanoseconds.

(not directly related to this PR which adds the file caching...)

rvps/src/cache/local_fs/mod.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    fn get(&self, name: &str) -> Result<Option<ReferenceValue>> {
+        match self.engine.get(name)? {
+            Some(v) => {
+                let v = serde_json::from_slice(&v)?;
+                Ok(Some(v))
+            }
+            None => Ok(None),
+        }
```
dubek, line 62: maybe replace the entire match with:

```rust
self.engine.get(name)?.map(|v| serde_json::from_slice(&v)?)
```

something similar can be done above in `set()`.

Up to you if it's clearer or you prefer the `match`.

dubek (commented):
I probably missed something in the original RVPS PR: we use Cache as a persistent storage for reference values? Usually the term "cache" is used for something that you store to improve performance (to avoid re-computation or fetching from slow disk/db); but the system should continue to function correctly when the cache is erased (maybe more slowly). I don't fully understand here if Cache represents a cache (to improve performance) or a simple persistent storage.

Some small comments below:

rvps/src/reference_value.rs
Xynnn007: You're right, thanks for catching this

rvps/src/cache/local_fs/mod.rs
Xynnn007: I got your idea, but Rust seems not able to support `?` in a `map` closure. I tried this, but failed

dubek: Oh, you're right (it won't work because `?` works only when you return a Result, but the closure doesn't). Thanks.

dubek:
> Sorry for the confusing name `Cache`. When we first designed this module, it aimed to store the RVs instead of calculating it every time users requested for it. Another name like `Storage` would be better, I'd like to know your opinion that if this is something like a bug, I'll open a new PR to fix it. Thanks!

No worries. I think that `Storage` indeed better reflects the usage.

I guess you can change it in another PR and then other people can voice their opinions as well.

jialez0 (approved):

rvps/src/cache/local_fs/README.md
```diff
@@ -0,0 +1,15 @@
+# Local File System Cache
```
sameo, line 1: I agree with @dubek here, `cache` is confusing.
Can we rename `src/cache` to `src/store`, as a first commit to that PR?

rvps/src/cache/local_fs/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+This is a simple Cache, which will store the
+Reference Values in a local file. Thus the data
+of the RVPS is persistent, and also safe when it
+comes to restart or crash.
+
+The underlying storage engine is `sled`, which
```
sameo, line 8: Have you considered using the [kv](https://docs.rs/kv/latest/kv/) crate instead?

rvps/src/reference_value.rs
```diff
@@ -75,13 +76,19 @@ fn default_version() -> String {
...
+    pub fn new() -> Result<Self> {
+        Ok(ReferenceValue {
             version: REFERENCE_VALUE_VERSION.into(),
             name: String::new(),
-            expired: Utc::now(),
+            expired: Utc::now()
+                .with_nanosecond(0)
+                .ok_or_else(|| anyhow!("set nanosecond failed."))?,
```
sameo, line 89: A generic comment on that commit: I do not think it's related to the PR it's included with. So please send a separate PR with only that commit, as a fix.

rvps/src/cache/local_fs/README.md
Xynnn007: I have roughly investigated some key-value storages. Sled is the most popular one, [kv](https://docs.rs/kv/latest/kv/) is not used by many people. I'm not sure which is better, but I once used sled to develop a Key-Value Storage service and it worked well. So I choosed sled here.

rvps/src/cache/local_fs/README.md
Xynnn007: Sure. I'm going to open an issue to track this problem

rvps/src/reference_value.rs
Xynnn007: Yea, thanks! I'm going to

rvps/src/store/mod.rs
```diff
@@ -9,13 +9,16 @@ use crate::reference_value::ReferenceValue;
...
+
 /// Interface of a `Store`.
 /// We only provide a simple instance here which implements
 /// Store. In more scenarios, RV should be stored in persistent
 /// storage, like database, file and so on. All of the mentioned
 /// forms will have the same interface as following.
 pub trait Store {
-    /// Store a reference value
+    /// Store a reference value. If the given `name` exists,
+    /// return the previous `Some<ReferenceValue>`, or return `None`
```
dubek, line 21:
```suggestion
    /// Store a reference value. If the given `name` exists,
    /// return the previous `Some<ReferenceValue>`, otherwise return `None`
```

rvps/src/store/local_fs/mod.rs
```diff
@@ -0,0 +1,146 @@
...
+    /// for [`LocalFs`].
+    #[test]
+    #[serial]
+    fn set_and_get() {
+        {
+            let mut store = LocalFs::default();
+            let rv = ReferenceValue::new().expect("create ReferenceValue failed.");
+            store
+                .set(KEY.to_owned(), rv.clone())
+                .expect("set rv failed.");
```
dubek, line 85: maybe add a check that `store.set(...)` returned `None`.

rvps/src/store/local_fs/mod.rs
```diff
@@ -0,0 +1,146 @@
...
+
+use std::path::Path;
+
+use anyhow::*;
+
+use crate::{ReferenceValue, Store};
+
+/// Local directory path to store the reference values,
+/// which is created by sled engine.
+const FILE_PATH: &str = "./reference_values";
```
dubek, line 16: Will this file be inside the `ATTESTATION_SERVER_WORKDIR` ?

dubek (commented):
Looks good;  a few comments/questions below.

rvps/src/store/mod.rs
Xynnn007: Thanks, fixed it

rvps/src/store/local_fs/mod.rs
Xynnn007: Thanks, added it!

rvps/src/store/local_fs/mod.rs
Xynnn007: Yes, agreed with you. Now the new PR of AS doesn't have a `ATTESTATION_SERVER_WORKDIR` const, so I suppose to use the previous `/opt/attestation-server` in RVPS

rvps/src/store/local_fs/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+    use super::{LocalFs, FILE_PATH};
+    const KEY: &str = "test1";
+
+    /// This test will test the `set` and `get` interface
+    /// for [`LocalFs`].
+    #[test]
+    #[serial]
+    fn set_and_get() {
+        {
+            let mut store = LocalFs::default();
```
dubek, line 81: running the tests will erase `/opt/attestation-server/reference_values` ...

maybe the test should use `let mut store = LocalFs::new("test_localfs_dir")` instead of the default? (and similarly when calling remove_all_dir).
(related to all the test cases.)

Xynnn007: Thanks for pointing out! I'll fix this bug

dubek (approved):
LGTM

Issue #19, opened 2022-10-13
Pluggable policy engine for attestation server


bpradipt:
Currently the attestation server is tightly couple with Open Policy Agent (OPA).
It would be useful to make it modular and enable usage of alternative policy engines like [kyverno](https://kyverno.io).

jialez0:
> Currently the attestation server is tightly couple with Open Policy Agent (OPA). It would be useful to make it modular and enable usage of alternative policy engines like [kyverno](https://kyverno.io).

@bpradipt It will be easy to realize the modularization of the policy engine (only need to do some additional encapsulation for the policy verification part), but we may need to redesign the message format of AS interacting with the policy engine to make it universal enough to support various policy engines. In the subsequent PR, perhaps we can achieve this.

jialez0:
@bpradipt Hi, after #24 be merged, I will raise a new PR to pluggable policy engine :)

bpradipt:
> @bpradipt Hi, after #24 be merged, I will raise a new PR to pluggable policy engine :)

awesome @jialez0 :-)

jialez0:
Hi, @bpradipt , I have raise the PR #25  to solve this issue. Now you can review it.

PR #25, opened 2022-10-24
Pluggable policy engine.

jialez0:
resolve: #19 

cc @bpradipt


src/policy_engine/opa/mod.rs
```diff
@@ -16,68 +22,58 @@ pub struct GoString {
...
+            p: reference.as_ptr() as *const i8,
+            n: reference.len() as isize,
+        };
+
+        let input_go = GoString {
+            p: input.as_ptr() as *const i8,
+            n: input.len() as isize,
+        };
 
-    #[test]
```
fitzthum, line 69: Removing the tests?

src/lib.rs
```diff
@@ -8,55 +8,50 @@ extern crate strum;
...
         })
     }
 
-    /// Get current policy.
-    pub fn policy(&self) -> String {
-        self.policy.clone()
-    }
-
-    /// Set customize policy.
-    pub fn set_policy(&mut self, policy: String) {
```
fitzthum, line 56: So how do we set the policies now?

jialez0: Because the policy files of different policy engines have different formats, they are now set through the policy file path. For example, for `opa`, users can configure policy in path `/opt/confidential-containers/attestation-service/opa/policy.rego`, and the policy content in this file will be read during verification.

jialez0: Because the policy files of different policy engines have different formats, they are now set through the policy file path. For example, for `opa`, users can configure policy in path `/opt/confidential-containers/attestation-service/opa/policy.rego`, and the policy content in this file will be read during verification.

src/config.rs
```diff
@@ -0,0 +1,48 @@
+use anyhow::{anyhow, Result};
+use serde::Deserialize;
+use std::convert::TryFrom;
+use std::fs::File;
+use std::path::{Path, PathBuf};
+
+/// Environment macro for Attestation Service work dir.
+const AS_WORK_DIR: &str = "AS_WORK_DIR";
+const DEFAULT_WORK_DIR: &str = "/opt/attestation-service";
```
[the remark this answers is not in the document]
jialez0: That sounds good.

jialez0:
@bpradipt Hi, does this looks good to you? We can try to get it merged :-)

PR #26, opened 2022-11-07
AS: Integrated RVPS

jialez0:


Xynnn007:
The Ci seemed unhappy. I'm going to fix this first

rvps/src/lib.rs
```diff
@@ -131,4 +137,19 @@ impl<T: Store> RVPSAPI for Core<T> {
             }
         }
     }
+
+    fn generate_reference_data(&self, claims: &str) -> Result<HashMap<String, Vec<String>>> {
```
Xynnn007, line 141: Hi, it seems that this function is to realize "giving a `<componant-name>:<evidence-digest-gathered-by-TEE>`, return the `<trusted-digests>` from RVPS". I think the logic placed here may break the modularition of RVPS. Two possible ways I think:
* Maybe this logic should be placed in AS, and something like
```rust
let mut data = HashMap::new();
for key in claims_map.keys() {
    if let Some(digest) = <rvps-obj>.get_digests(key).await? {
        data.insert(key.clone(), digest);
    } else {
        anyhow::bail!("Can not get a trust value from RVPS for {key}");
    }
    
}
```
* Change the API of RVPS `get_digests`. From `fn get_digests(&self, name: &str) -> Result<Option<TrustedDigest>>` to `fn get_digests(&self, name: &[&str]) -> Result<Vec<Option<TrustedDigest>>>;` to support request for multiple `TrustDigest`s in one request.

How do you like?

rvps/src/store/local_fs/mod.rs
```diff
@@ -13,7 +13,7 @@ use crate::{ReferenceValue, Store};
 
 /// Local directory path to store the reference values,
 /// which is created by sled engine.
-const FILE_PATH: &str = "/opt/attestation-server/reference_values";
+const FILE_PATH: &str = "/opt/confidential-containers/attestation-service/reference_values";
```
Xynnn007, line 16: Do we need to give a document to record all the paths that CoCo uses in all sub-projects? (e.g., AS, AA, image-rs, etc) I think it may help develop and test a lot

jialez0: Good idea.

Xynnn007 (approved):
LGTM, thanks!

jiazhang0 (approved):

PR #32, opened 2022-11-24
Feat in toto handler

Xynnn007:
### Summary

For https://github.com/confidential-containers/attestation-service/issues/23, this pr implements in-toto handler with a rust wrapper for golang-in-toto.

### Change log

* Refactored RVPS into a submodule of AS, which helps to collect all the cgo code into one `build.rs` file to prevent errors when linking. (This modified including paths and dependencies in `Cargo.toml`)
* Added handler for in-toto, which will call a rust-c-go wrapper (under dir `shim`) into golang-in-toto's verifylib
* Added a unit test for in-toto's handler. Some test in-toto metadata files were added in `tests/in-toto`
* Added in-toto as a default feature
* Combined all cgo modules (opa and in-toto-golang) into one static library (go package, so that resulted `cgo.a`) to avoid the error of duplicated definition when rust links two different go-built static libraries. This makes `go.mod` and `go.sum` changed.
* Reordered all the dependencies in `Cargo.toml` lexicographically.

### Related Issues
Close https://github.com/confidential-containers/attestation-service/issues/34
Close https://github.com/confidential-containers/attestation-service/issues/23


Xynnn007:
A quick question, shall I need to add the code of the Rust wrapper for golang-in-toto in this? Or just import that in `Cargo.toml`?

jialez0 (approved):
LGTM.

PR #33, opened 2022-11-29
AS: Add TDX Verifier

jialez0:
resolve: #30

Verify the evidence generated by https://github.com/confidential-containers/attestation-agent/pull/110

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,6 @@ path = "bin/grpc-as/src/main.rs"
 [features]
 default = [ "in-toto" ]
 in-toto = []
-sample_verifier = []
 
 [dependencies]
 anyhow = "1.0"
@@ -23,7 +22,7 @@ futures = "0.3.17"
 # TODO: Replace this with crate.io published version
 in-toto = { git = "https://github.com/Xynnn007/in-toto-rs.git", rev = "7f69799" }
 # TODO: Replace this with kbs-types next published version (0.2.0)
-kbs-types = { git = "https://github.com/virtee/kbs-types.git" }
+kbs-types = { git = "https://github.com/virtee/kbs-types.git", rev = "50ec211" }
 lazy_static = "1.4.0"
 log = "0.4.17"
 path-clean = "0.1.0"
--- a/src/verifier/mod.rs
+++ b/src/verifier/mod.rs
@@ -5,15 +5,10 @@ use kbs_types::{Attestation, Tee};
 
 pub mod sample;
 
-#[cfg(feature = "sample_verifier")]
-pub(crate) fn to_verifier(_tee: &Tee) -> Result<Box<dyn Verifier + Send + Sync>> {
-    Ok(Box::new(sample::Sample::default()) as Box<dyn Verifier + Send + Sync>)
-}
-
-#[cfg(not(feature = "sample_verifier"))]
 pub(crate) fn to_verifier(tee: &Tee) -> Result<Box<dyn Verifier + Send + Sync>> {
     match tee {
         Tee::Sev | Tee::Sgx | Tee::Snp | Tee::Tdx => todo!(),
+        Tee::Sample => Ok(Box::<sample::Sample>::default() as Box<dyn Verifier + Send + Sync>),
     }
 }
 


Xynnn007:
The use of `ring` might cause some problems with IBM `s390x` arch. I suggest to replace it with RustCrypto in this pr or another. If another, I'd like to help.

Cargo.toml
```diff
@@ -15,19 +15,24 @@ in-toto = []
...
 lazy_static = "1.4.0"
 log = "0.4.17"
 path-clean = "0.1.0"
 prost = "0.11.0"
 redis = "0.21.5"
+ring = { git = "https://github.com/briansmith/ring.git", default-features = false, features = ["alloc"]}
+scroll = { version = "0.11.0", default-features = false, features = ["derive"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "*"
+sgx-dcap-quoteverify-rs = { git = "https://github.com/intel/SGXDataCenterAttestationPrimitives" }
```
Xynnn007, line 35: Should we specify the rev `85cf8bdd`, which can prevent API change.

src/verifier/tdx/eventlog.rs
```diff
@@ -0,0 +1,238 @@
...
+            rtmr1: rtmr1[0..48].try_into()?,
+            rtmr2: rtmr2[0..48].try_into()?,
+            rtmr3: rtmr3[0..48].try_into()?,
+        };
+
+        Ok(mr)
+    }
+
+    pub fn query_digest(&self, entity: MeasuredEntity) -> Option<String> {
+        let event_desc_prefix = Self::generate_quary_key_prefix(entity)?;
```
Xynnn007, line 118: `quary` -> `query`

src/verifier/tdx/mod.rs
```diff
@@ -0,0 +1,164 @@
...
+                Some(td_paylod_info_digest) => {
+                    claim_map.insert(
+                        "tdx-kernel-params".to_string(),
+                        serde_json::Value::String(td_paylod_info_digest),
+                    );
+                }
+                None => todo!(),
+            }
+        }
+        None => todo!(),
```
Xynnn007, line 158: Do we need to report an error here?

src/verifier/tdx/mod.rs
```diff
@@ -0,0 +1,164 @@
...
+                None => todo!(),
+            }
+            match ccel.query_digest(MeasuredEntity::TdPayloadInfo) {
+                Some(td_paylod_info_digest) => {
+                    claim_map.insert(
+                        "tdx-kernel-params".to_string(),
+                        serde_json::Value::String(td_paylod_info_digest),
+                    );
+                }
+                None => todo!(),
```
Xynnn007, line 155: Do we need to report an error here?

src/verifier/tdx/eventlog.rs
```diff
@@ -0,0 +1,238 @@
...
+    pub fn query_digest(&self, entity: MeasuredEntity) -> Option<String> {
+        let event_desc_prefix = Self::generate_query_key_prefix(entity)?;
+
+        for (header, data) in self.cc_events {
+            if data.len() < event_desc_prefix.len() {
+                continue;
+            }
+            if &data[..event_desc_prefix.len()] == event_desc_prefix.as_slice() {
+                let digest = &header.digest.digests[0];
+                let sha384 = digest.digest.sha384;
```
Xynnn007, line 126: To make it robust, do we need to assert `digest.hash_alg == TPML_ALG_SHA384`?

src/verifier/tdx/mod.rs
```diff
@@ -0,0 +1,170 @@
...
+    );
+    claim_map.insert(
+        "tdx-mrconfigid".to_string(),
+        serde_json::Value::String(hex::encode(quote.report_body.mr_config_id)),
+    );
+
+    // Claims from CC EventLog.
+    match cc_eventlog {
+        Some(ccel) => {
+            match ccel.query_digest(MeasuredEntity::TdPayload) {
```
Xynnn007, line 139: Some question from my point of view. Let's take kernel for example. Here the query is based on a principle that the desired CcEvent of the entity will have `11 | "td_payload\0"` as the first bytes of the `data` of the `CcEvent`.

However, when the length of the hashed content of the kernel (s.t. `payload` in td-shim) varies, the expected measurement should also vary. Thus I'd suggest that we add more information about the measured entity, s.t. we replace the key here
```
"tdx-kernel": "<sha384-of-the-evidence>"
```
with 
```
"tdx-kernel": 
    {
    "value": "<sha384-of-the-evidence>",
    "size": "<KERNEL_SIZE-from-CcEvent-data>"
    }
```
or 
```
"tdx-kernel-<KERNEL_SIZE-from-CcEvent-data>": "<sha384-of-the-evidence>"
```

to help get reference data from rvps when the size is different.

src/verifier/tdx/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+    claim_map.insert(
+        "tdx-mrconfigid".to_string(),
+        serde_json::Value::String(hex::encode(quote.report_body.mr_config_id)),
+    );
+
+    // Claims from CC EventLog.
+    match cc_eventlog {
+        Some(ccel) => {
+            match (
+                ccel.query_digest(MeasuredEntity::TdPayload),
```
Xynnn007, line 140: Hi, do we need to measure Kernel-Parameter? I've roughly thought that we can directly measure kernel-para using a simple digest/reference value as mentioned here https://github.com/confidential-containers/attestation-service/pull/43/files#diff-633920d19cd2b94035233ca8d93f72637166955663f94369bcfd5927e00fd441R22 . However, CCEL provides the plaintext of the kernel-para. 

I'm not sure which option is better:
- Measure the para by matching a reference digest. The code logic will be easier
- Parse the parameters and split them into a vec, then apply some policy. More flexible, but needs more coding and design

PR #38, opened 2022-12-07
SNP Verifier

fitzthum:
I think the signature format might be wrong, but might as well post this so we can think abou it as the other pieces start to emerge.

Signed-off-by: Tobin Feldman-Fitzthum <tobin@ibm.com>

@dubek


larrydewey:
This is more for documentation sake, but like we talked about this on one of our calls, unfortunately both`{generation}::ARK|ASK` are in the [legacy SEV certificate format](https://www.amd.com/system/files/TechDocs/55766_SEV-KM_API_Specification.pdf#page=118). Currently virtee/sev only offers support for the legacy format (required for SEV/SEV-ES), but they have plans to add support for the x509 certificate format in the future. The x509 certificates may be pulled from the following website:

https://kdsintf.amd.com/vcek/v1/{product_name}/cert_chain

Where `produce_name` is capitalized. This will return a certificate chain in PEM format.

src/verifier/snp/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+use sha2::{Digest, Sha384};
+
+#[derive(Serialize, Deserialize)]
+struct SnpEvidence {
+    attestation_report: AttestationReport,
+    cert_chain: CertTable,
+}
+
+#[derive(Debug, Default)]
+pub struct Snp {}
```
larrydewey, line 25: I am not seeing this used anywhere. Is this just a place-holder for the moment, or just to tie the impl to a generic?

src/verifier/snp/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    sig.verify(&data, EcKey::try_from(vcek.public_key()?)?.as_ref())
+        .map_err(|_| anyhow!("Signature validation failed."))?;
+
+    Ok(())
+}
+
+fn verify_cert_chain(cert_chain: &CertTable) -> Result<x509::X509> {
+    // add support for genoa
+    let certs = x509::X509::stack_from_pem(include_bytes!("milan_ark_ask.pem"))?;
```
larrydewey, line 73: Given the rare possibility of a key revocation, I am assuming we would just update the file locally?

src/verifier/snp/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+    let ask = &certs[0];
+    let ark = &certs[1];
+
+    // check for revocation?
+
+    //let raw_vcek = cert_chain.entries.iter().find(|c| c.cert_type == SnpCertType::VCEK)
+    let raw_vcek = cert_chain
+        .entries
+        .iter()
+        .find(|c| c.guid() == "63da758d-e664-4564-adc5-f4b93be8accd")
```
larrydewey, line 85: I don't believe the VCEK is part of the certificate chain you are referencing above. So this would almost always fail.

src/verifier/snp/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+        .ok_or(anyhow!("VCEK not found."))?;
+    let vcek =
+        x509::X509::from_der(&raw_vcek.data()).map_err(|_| anyhow!("Failed to load VCEK"))?;
+
```
larrydewey, line 89: Would it make sense to verify the ARK is self-signed?

src/verifier/snp/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+        // policy fields
+        "policy_abi_major": report.policy.abi_major(),
+        "policy_abi_minor": report.policy.api_minor(),
+        "policy_smt_allowed": report.policy.smt_allowed(),
+        "policy_migrate_ma": report.policy.migrate_ma_allowed(),
+        "policy_debug_allowed": report.policy.debug_allowed(),
+        "policy_single_socket": report.policy.single_socket_required(),
+
+        // versioning info
+        "current_tcb_bootloader": report.current_tcb.boot_loader,
```
larrydewey, line 124: I think the more important one to check is actually `reported_tcb`.

larrydewey (changes requested):
Looking good so far, just a couple of things I noticed. Also, do you have plans to verify the VCEK  x509v3 fields with the tcb content pulled from the attestation report?

src/verifier/snp/mod.rs
fitzthum: I think we're expecting to VCEK to be provisioned at operator deployment time and to be part of he cert table in the extended report. Does that seem right?

src/verifier/snp/mod.rs
fitzthum: I think we need to compare the VCEK to the reported TCB. I forgot to implement this. I am still making up my mind about which tcb version should actually be used as a claim.

src/verifier/snp/mod.rs
larrydewey: Ah, `certs` vs `cert_chain` my mistake.

fitzthum: I'm going to clean up the parsing of the certs file and that should make it a bit more clear.

src/verifier/snp/milan_ark_ask.pem
```diff
@@ -0,0 +1,74 @@
+-----BEGIN CERTIFICATE-----
```
dubek, line 1: if ASK is first in this file, then maybe rename it to `milan_ask_ark.pem` ?

src/verifier/snp/mod.rs
dubek: also: extract the guid string to a const and add a comment referencing the GHCB spec section.

fitzthum: I'm going to ditch this guid once he cert struct has the cert type as public.

src/verifier/snp/milan_ark_ask.pem
fitzthum: Turns out the ARK is first after all.

src/verifier/snp/mod.rs
```diff
@@ -0,0 +1,170 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use openssl::nid::Nid;
+
+    #[test]
+    fn check_milan_certificates() {
+        let (ark, ask) = load_milan_cert_chain().unwrap();
+        assert_eq!(get_common_name(ark).unwrap(), "SEV-Milan");
+        assert_eq!(get_common_name(ask).unwrap(), "ARK-Milan");
```
dubek, line 156: So I split the pem file into 1.pem (first) and 2.pem (second) certs. Then I looked at them:

```
$ openssl x509 -in 2.pem -text -noout
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number: 65536 (0x10000)
        Signature Algorithm: rsassaPss
         Hash Algorithm: sha384
         Mask Algorithm: mgf1 with sha384
          Salt Length: 0x30
         Trailer Field: 0x01
        Issuer: OU = Engineering, C = US, L = Santa Clara, ST = CA, O = Advanced Micro Devices, CN = ARK-Milan
        Validity
            Not Before: Oct 22 17:23:05 2020 GMT
            Not After : Oct 22 17:23:05 2045 GMT
        Subject: OU = Engineering, C = US, L = Santa Clara, ST = CA, O = Advanced Micro Devices, CN = ARK-Milan
```

this (2.pem) looks like a self-signed cert (issuer = subject), and its subject is `ARK-Milan`.

as for the first one:

```
$ openssl x509 -in 1.pem -text -noout
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number: 65537 (0x10001)
        Signature Algorithm: rsassaPss
         Hash Algorithm: sha384
         Mask Algorithm: mgf1 with sha384
          Salt Length: 0x30
         Trailer Field: 0x01
        Issuer: OU = Engineering, C = US, L = Santa Clara, ST = CA, O = Advanced Micro Devices, CN = ARK-Milan
        Validity
            Not Before: Oct 22 18:24:20 2020 GMT
            Not After : Oct 22 18:24:20 2045 GMT
        Subject: OU = Engineering, C = US, L = Santa Clara, ST = CA, O = Advanced Micro Devices, CN = SEV-Milan
```

This is signed by the ARK, so this looks to me like the ASK (even though the subject doesn't say ASK).

So are you sure about the order of the certs in the pem file?

src/verifier/snp/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+        .map_err(|_| anyhow!("Invalid VCEK Signature"))?;
+
+    Ok(vcek)
+}
+
+fn calculate_expected_report_data(nonce: &String, tee_pubkey: &String) -> [u8; 64] {
+    let mut hasher = Sha384::new();
+
+    hasher.update(nonce.clone().into_bytes());
+    hasher.update(tee_pubkey.clone().into_bytes());
```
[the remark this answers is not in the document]
dubek: OK. Then keep the strings, and replace `clone()` with `as_bytes()` ?

src/verifier/snp/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+    // check cert chain
+    let vcek = verify_cert_chain(&evidence.cert_chain)?;
+
+    // verify report signature
+    let sig = ecdsa::EcdsaSig::try_from(&evidence.attestation_report.signature)?;
+    let data = &bincode::serialize(&evidence.attestation_report)?[..=0x29];
+
+    sig.verify(&data, EcKey::try_from(vcek.public_key()?)?.as_ref())
+        .map_err(|_| anyhow!("Signature validation failed."))?;
+
```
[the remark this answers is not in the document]
fitzthum: Added

fitzthum:
Added in some code to check the chip ID and tcb version. Still needs a bit of cleanup. I will add tests with a VCEK shortly.

src/verifier/snp/mod.rs
```diff
@@ -0,0 +1,265 @@
...
+    #[test]
+    fn check_vcek_signature() {
+        let (ask, _ark) = load_milan_cert_chain().unwrap();
+        let vcek = x509::X509::from_der(include_bytes!("test-vcek.der")).unwrap();
+
+        // ASK -> VCEK
+        vcek.verify(&(ask.public_key().unwrap() as PKey<Public>))
+            .map_err(|_| anyhow!("Invalid VCEK Signature"))
+            .unwrap();
+    }
```
dubek, line 264: 1. I suggest modifying this to use `verify_cert_chain()` which is the code that actually runs in the verifier:

```rust
#[cfg(test)]
mod tests {
...
    use sev::firmware::host::types::CertTableEntry;
...
    #[test]
    fn check_vcek_signature() {
        let vcek = include_bytes!("test-vcek.der").to_vec();
        let cert_table: CertTable = CertTable {
            entries: vec![
                CertTableEntry::new(SnpCertType::VCEK, vcek),
            ],
        };
        verify_cert_chain(&cert_table).unwrap();
    }
}
```

2. I suggest adding a negative check -- somehow corrupting test-vcek and checking that `verify_cert_chain()` returns an Err.

fitzthum: Added

dubek (approved):
Looks good!

src/verifier/snp/milan_ark_ask.pem
larrydewey: Actually, the ASK is first @fitzthum (at least as of Feb 7, 2023).

src/verifier/snp/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+
+    val.try_into()
+        .map_err(|_| anyhow!("Oid data has wrong length."))
+}
+
+fn get_oid_int(
+    vcek: &x509_parser::certificate::TbsCertificate,
+    oid: Oid,
+) -> Result<u8> {
+    let val_raw = get_oid::<3>(&vcek, oid)?;
```
larrydewey, line 81: Do we know that the raw value will always be 3 bytes? What about the ucode, can't that be 4 bytes in size?

src/verifier/snp/mod.rs
larrydewey: @fitzthum as of `1.1.0` you may now ditch this.

src/verifier/snp/mod.rs
fitzthum: I think that's why I am getting 3 bytes. Originally I was trying to directly read the bytes, but this gave me the wrong results. Instead I get three bytes and give it to the der parser, which I think will handle this padding thing you mention?

fitzthum: This should be fixed now btw.

src/verifier/snp/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+    let mut hash = hasher.finalize().to_vec();
+    hash.extend([0u8; 16]);
+
+    hash.as_slice().try_into().unwrap()
+}
+
+fn parse_tee_evidence(report: &AttestationReport) -> Result<TeeEvidenceParsedClaim> {
+    let claims_map = json!({
+        // policy fields
+        "policy_abi_major": report.policy.abi_major(),
```
mkulke, line 183: I think those need to be string => string maps (see [here](https://github.com/confidential-containers/attestation-service/blob/main/src/lib.rs#L131)). It's maybe a bit unfortunate that the parsed claim is a untyped json value, since we deserialize the string in the same crate

src/verifier/snp/mod.rs
```diff
@@ -0,0 +1,273 @@
...
+    hasher.update(nonce.as_bytes());
+    hasher.update(tee_pubkey.as_bytes());
+
+    let mut hash = hasher.finalize().to_vec();
+    hash.extend([0u8; 16]);
+
+    hash.as_slice().try_into().unwrap()
+}
+
+fn parse_tee_evidence(report: &AttestationReport) -> Result<TeeEvidenceParsedClaim> {
```
mkulke, line 180: Shouldn't return a Result I think

fitzthum: Fair enough

src/verifier/snp/mod.rs
fitzthum: Hm I added in some formats. Not sure this is really optimal, but I will double check the values when we test the evidence parsing. Haven't quite gotten to that point yet.

fitzthum:
@mkulke @larrydewey Have your comments on this been addressed? I would like to merge soonish

src/verifier/snp/mod.rs
```diff
@@ -0,0 +1,274 @@
...
+    let mut hasher = Sha384::new();
+
+    hasher.update(nonce.as_bytes());
+    hasher.update(&tee_pubkey.k_mod);
+    hasher.update(&tee_pubkey.k_exp);
+
+    let mut hash = hasher.finalize().to_vec();
+    hash.extend([0u8; 16]);
+
+    hash.as_slice().try_into().unwrap()
```
mkulke, line 179: nit: if you want to get rid of the result/unwrap you should be able to copy the slice content directly:

```suggestion
    let partial_hash = hasher.finalize();

    let mut hash = [0u8; 64];
    hash[..48].copy_from_slice(&partial_hash);
    hash
```

mkulke (approved):
nit, otherwise lgtm

src/verifier/snp/mod.rs
fitzthum: Good suggestion. Thanks

fitzthum:
@jiale @sameo PTAL

attestation-service/src/verifier/snp/mod.rs
```diff
@@ -0,0 +1,272 @@
...
+        .context("Signature validation failed.")?;
+
+    Ok(())
+}
+
+fn load_milan_cert_chain() -> Result<(x509::X509, x509::X509)> {
+    let certs = x509::X509::stack_from_pem(include_bytes!("milan_ask_ark.pem"))?;
+
+    // ask, ark
+    Ok((certs[0].clone(), certs[1].clone()))
```
sameo, line 138: Nit, we don't know if there are actually 2 certs in that chain.

sameo (commented):
Just a couple nits, overall lgtm.

attestation-service/src/verifier/snp/mod.rs
fitzthum: The cert we are reading is tracked along with the source, so we can be pretty sure of its format. Should I add a check for the length?

attestation-service/src/verifier/snp/mod.rs
```diff
@@ -0,0 +1,272 @@
...
+
+    // ASK -> VCEK
+    vcek.verify(&(ask.public_key()? as PKey<Public>))
+        .context("Invalid VCEK Signature")?;
+
+    Ok(vcek)
+}
+
+fn calculate_expected_report_data(nonce: &String, tee_pubkey: &TeePubKey) -> [u8; 64] {
+    let mut hasher = Sha384::new();
```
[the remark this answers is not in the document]
fitzthum: This corresponds to [this](https://github.com/confidential-containers/attestation-agent/blob/main/deps/crypto/src/teekey.rs#L63). Ideally we should extract this hashing from all the verifiers into some utility function or even use the one from the AA, but that might be best done in another PR.

attestation-service/src/verifier/snp/mod.rs
sameo: > The cert we are reading is tracked along with the source, so we can be pretty sure of its format. Should I add a check for the length?

I think so, yes.

sameo (approved):

PR #39, opened 2022-12-08
kbs-types sync

sameo:
This PR syncs with the[`kbs-types`](https://github.com/virtee/kbs-types) in order to have common KBS type definitions (`Tee`, `Attestation`) for both the AS and the KBS implementations.

One consequence of using the `kbs-types` crate is that we can no longer easily instantiate the `Sample` verifier. This PR puts it under a feature flag (`sample_verifier`). When turned on, the only available verifier is the `Sample` one. When turned off, `Sample` is not available.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ path = "bin/grpc-as/src/main.rs"
 [features]
 default = [ "in-toto" ]
 in-toto = []
+sample_verifier = []
 
 [dependencies]
 anyhow = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,8 @@ env_logger = "0.9.1"
 futures = "0.3.17"
 # TODO: Replace this with crate.io published version
 in-toto = { git = "https://github.com/Xynnn007/in-toto-rs.git", rev = "7f69799" }
+# TODO: Replace this with kbs-types next published version (0.2.0)
+kbs-types = { git = "https://github.com/virtee/kbs-types.git" }
 lazy_static = "1.4.0"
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -21,6 +22,8 @@ env_logger = "0.9.1"
 futures = "0.3.17"
 # TODO: Replace this with crate.io published version
 in-toto = { git = "https://github.com/Xynnn007/in-toto-rs.git", rev = "7f69799" }
+# TODO: Replace this with crate.io published version
+kbs-types = { git = "https://github.com/sameo/kbs-types.git", branch = "topic/kbs" }
```
sameo, line 26: I will switch to the upstream branch once https://github.com/virtee/kbs-types/pull/11 is merged.

jialez0 (approved):
LGTM.

Cargo.toml
sameo: The kbs-types PR is merged now, and I updated that PR to point to upstream.

dubek:
(I know this was already merged.)

@sameo This looks good, but the current CI doesn't build/test with the new `sample_verifier` feature that was introduced. To verify that, I changed a part in src/verifier/mod.rs to this:

```rust
#[cfg(feature = "sample_verifier")]
pub(crate) fn to_verifier(_tee: &Tee) -> Result<Box<dyn Verifier + Send + Sync>> {
    jfkljadskjflds;
    Ok(Box::new(sample::Sample::default()) as Box<dyn Verifier + Send + Sync>)
}
```

and `make` still completed successfully without an issue.

PR #47, opened 2023-01-29
[RVPS] Integrate in AS-gRPC, RVPS binary and documents

Xynnn007:
This PR include RVPS-related things:
- Add RVPS binary
- Let AS support two ways to interact with RVPS, s.t. integrate inside and by gRPC. This can be controlled by rust feature
- Document about RVPS


Xynnn007:
The error is caused by the upstream change (virtee-kbs types), I'll open another pr to fix this.

Cargo.toml
```diff
@@ -6,10 +6,19 @@ edition = "2021"
...
+[[bin]]
+name = "rvps"
+path = "bin/rvps/src/main.rs"
+required-features = [ "rvps-server", "rvps-proxy", "tokio/rt-multi-thread" ]
 
 [features]
-default = [ "in-toto" ]
-in-toto = []
+default = [ "rvps-server" ]
+rvps-server = [ "in-toto" ]
```
jialez0, line 18: Why not directly use `"dep:in-toto` here, so that `in-toto` feature can be remolved below.

Xynnn007: My initial idea is to make `in-toto` as a option feature. Here I made a mistake, let me fix into
```
default = [ "rvps-server" ]
rvps-server = []
rvps-proxy = [ "tonic" ]

in-toto = [ "dep:in-toto" ]
```

In this way, we can choose whether to enable `in-toto` support

jialez0:
@Xynnn007 Thanks! Could you please update the architecture diagram in README.md to follow this PR?

jialez0 (approved):
LGTM. Thanks.

PR #49, opened 2023-01-30
AS: Add TDX Verifier

jialez0:
This PR inherits #33 (closed by mistake and cannot be reopened due to branch failure)


src/verifier/tdx/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+    claim_map.insert(
+        "tdx-mrconfigid".to_string(),
+        serde_json::Value::String(hex::encode(quote.report_body.mr_config_id)),
+    );
+
+    // Claims from CC EventLog.
+    match cc_eventlog {
+        Some(ccel) => {
+            match (
+                ccel.query_digest(MeasuredEntity::TdPayload),
```
Xynnn007, line 140: Same as mentioned, do we need to measure Kernel-Parameter? I've roughly thought that we can directly measure kernel-para using a simple digest/reference value as mentioned here https://github.com/confidential-containers/attestation-service/pull/43/files#diff-633920d19cd2b94035233ca8d93f72637166955663f94369bcfd5927e00fd441R22 . However, CCEL provides the plaintext of the kernel-para. 

I'm not sure which option is better:
- Measure the para by matching a reference digest. The code logic will be easier. But any reordering of the same paras will result in different measurement.
- Parse the parameters and split them into a vec, then apply some policy. More flexible, but needs more coding and design

src/verifier/tdx/quote.rs
```diff
@@ -0,0 +1,255 @@
...
+        .try_into()
+        .unwrap();
+
+    let p_supplemental_data = match supplemental_data_size {
+        0 => None,
+        _ => Some(&mut supplemental_data),
+    };
+
+    // Call DCAP quote verify library for quote verification
+    let dcap_ret = qvl::sgx_qv_verify_quote(
```
Xynnn007, line 187: I have checked the source code inside DCAP. SGX and TDX share a same entrance for quote verification. :D We can support SGX then with less effort

src/verifier/tdx/mod.rs
jialez0: The second solution looks more beautiful. It needs some development work. Now this PR is large enough. After this PR is merged, let us implement it in a new PR.

Xynnn007 (approved):
LGTM, great work! Thanks!
Let's add some CI tests and check if they work correctly then

src/verifier/tdx/quote.rs
```diff
@@ -0,0 +1,253 @@
...
+
+pub fn parse_tdx_quote(quote_bin: &[u8]) -> Result<Quote> {
+    let quote_body = &quote_bin[..QUOTE_PAYLOAD_SIZE];
+    quote_body
+        .pread::<Quote>(0)
+        .map_err(|e| anyhow!("Parse TD quote failed: {:?}", e))
+}
+
+pub async fn ecdsa_quote_verification(quote: &[u8]) -> Result<()> {
+    let mut supplemental_data_size = 0u32; // mem::zeroed() is safe as long as the struct doesn't have zero-invalid types, like pointers
```
sameo, line 150: Nit: This comment applies to the next line.

sameo (approved):
Just one nit and one question. Otherwise, that's a pretty clean PR. Thanks @jialez0 and @Xynnn007 !

Cargo.toml
```diff
@@ -36,15 +39,19 @@ lazy_static = "1.4.0"
 log = "0.4.17"
 path-clean = "0.1.0"
 prost = "0.11.0"
+redis = "0.21.5"
```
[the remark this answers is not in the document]
jialez0: This should be an oversight in development. Removed now, thanks!

chendave:
Can we avoid the TDX dependencies when build the binary for AS or KBS, see: https://github.com/confidential-containers/attestation-service/issues/55

PR #50, opened 2023-02-15
Fix `make` && `make install` in the top dir

chendave:
It was:

> for bin_name in grpc-as grpc-as-ctl; do \
        install -D -m0755 target/release/$bin_name /usr/local/bin; \
done
install: cannot stat 'target/release/grpc-as': No such file or directory
install: cannot stat 'target/release/grpc-as-ctl': No such file or directory
Makefile:23: recipe for target 'install' failed

Now it should be okay,

>    Compiling attestation-service v0.1.0 (/home/dave/tmp/attestation-service)
    Finished release [optimized] target(s) in 1m 58s
for bin_name in grpc-as; do \
        install -D -m0755 target/release/$bin_name /usr/local/bin; \
done

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,3 +25,6 @@ install:
 	for bin_name in $(BIN_NAMES); do \
 		install -D -m0755 $(TARGET_DIR)/$$bin_name $(DESTDIR); \
 	done
+
+clean:
+	cargo clean
--- a/Makefile
+++ b/Makefile
@@ -16,8 +16,10 @@ else
     TARGET_DIR := $(TARGET_DIR)/release
 endif
 
-build:
-	cargo build $(release)
+build: grpc-as
+
+grpc-as:
+	cargo build --bin grpc-as --features rvps-server,rvps-proxy,tokio/rt-multi-thread $(release)
 
 install:
 	for bin_name in $(BIN_NAMES); do \
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ TOP_DIR := .
 CUR_DIR := $(shell pwd)
 PREFIX := /usr/local
 TARGET_DIR := target
-BIN_NAMES := grpc-as grpc-as-ctl
+BIN_NAMES := grpc-as
 
 DEBUG ?=
 DESTDIR ?= $(PREFIX)/bin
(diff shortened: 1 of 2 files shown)


chendave:
cc @jialez0

Cargo.toml
```diff
@@ -14,7 +14,7 @@ path = "bin/rvps/src/main.rs"
 required-features = [ "rvps-server", "rvps-proxy", "tokio/rt-multi-thread" ]
 
 [features]
-default = [ "rvps-server", "in-toto" ]
+default = [ "rvps-server", "in-toto", "rvps-proxy", "tokio/rt-multi-thread" ]
```
Xynnn007, line 17: By default we do not need `rvps-proxy` and `tokio/rt-multi-thread` feature to be enabled.
My naive idea is to replace
```bash
cargo build $(release)
```
in line 20 of `Makefile`
```bash
cargo build --bin grpc-as --features rvps-server,rvps-proxy,tokio/rt-multi-thread
```

How do you like?

Xynnn007 (commented):
Thanks for the contribution @chendave ! Some comments, please feel free to share your idea

Cargo.toml
chendave: If there multiple binaries are expected to be built, like it was `grpc-as grpc-as-ctl`, then cargo will repeat the command to build each of them with the enabled features.

chendave: I thought about this

```bash
cargo build --bin grpc-as --features rvps-server,rvps-proxy,tokio/rt-multi-thread
```

Just think it might not flexible, but I am fine with the approach as well.

Xynnn007: `cargo build` will only build the lib `attestation-service`. 
- `tokio/rt-multi-thread` is only used when an outer caller of the lib wants to use `tokio::main` or some other runtimes.
- `rvps-server` and `rvps-proxy` usually will not be used at the same time, excepte for `grpc-bin`.
For the two reasons, a specific build command would be better, I think.

> If there multiple binaries are expected to be built, like it was `grpc-as grpc-as-ctl`, then cargo will repeat the command to build each of them with the enabled features.

Yes. Some binaries might not depend all these features of AS, which will help to reduce the resulted binary size.

jialez0: It seems that it is better for us to regard the compilation of each binary as a make option? Something like:

```
grpc-as:
    cargo build --bin grpc-as --features rvps-server,rvps-proxy,tokio/rt-multi-thread
```

So that we can build `grpc-as` with:

```
make grpc-as
```

chendave: +1, updated, thanks for the suggestion.

chendave:
btw, is it possible for me to trigger the CI? seems like those jobs hasn't run since last push.

jialez0 (approved):
LGTM. Thanks! @chendave

PR #59, opened 2023-03-09
RVPS | Client for rvps

Xynnn007:
Added a simple client tool to register & query reference values to & from RVPS.

I test it locally and fixed some related bugs.

### DIFF
--- a/bin/rvps/README.md
+++ b/bin/rvps/README.md
@@ -1,0 +1,6 @@
+# Reference Value Provider Service
+
+This is the binary of RVPS
+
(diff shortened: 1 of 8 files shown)


Xynnn007:
Close https://github.com/confidential-containers/attestation-service/issues/60

jialez0:
@Xynnn007 Do you think we should put this client tool in a separate directory like `./tools/rvps-client` or some else?

jialez0 (approved):
LGTM.

PR #65, opened 2023-03-13
Fixed lacking of RVPS server of grpc-as and support specify config path.

jialez0:


Xynnn007 (approved):
This is pretty good for me! Thanks a lot for making up the config delivery!

src/lib.rs
```diff
@@ -67,8 +72,12 @@ impl AttestationService {
 
     /// Create a new Attestation Service, and connect to a remote rvps.
     #[cfg(feature = "rvps-proxy")]
-    pub async fn new_with_rvps_proxy(rvps_addr: &str) -> Result<Self> {
-        let _config = Config::default();
+    pub async fn new_with_rvps_proxy(rvps_addr: &str, config_path: Option<&str>) -> Result<Self> {
```
Xynnn007, line 75: A little question: Why not we pass `Config` struct here? Sometimes a user might want to change the config in  the code

jialez0: Good suggestion, I have changed it.

PR #78, opened 2023-03-27
Update kbs-types to v0.2.0

### LINKED ISSUE #77: Update kbs-types crate to v0.2.0
At the moment we reference the git repo of `kbs-types` instead of the published crate. This is seemingly due to a breaking change in v0.2.0, which instead of a pem string puts modulus and exponent in the `TeePubKey`.

mkulke:
The pubkey logic needs to be adjusted to match the upstream virtee/kbs-types structs. The fix needs to be coupled with the respective PRs on [Attestation Agent](https://github.com/confidential-containers/attestation-agent/pull/170) and [KBS](https://github.com/confidential-containers/attestation-agent/pull/170).

fixes #77


Xynnn007 (approved):
LGTM. And we should work together with AA side to test

mkulke:
> LGTM. And we should work together with AA side to test

Sure, is there a kbs channel on slack, so we can coordinate?

sameo (approved):

PR #79, opened 2023-03-28
Fix for pointer type error

tnakaike:
Pointer type conversion fails on s390x [here](https://github.com/confidential-containers/attestation-service/blob/0f7a70e98fc14890f2842dc66222b55ff28e1d99/src/policy_engine/opa/mod.rs#L59). This is a [known issue](https://github.com/rust-lang/rust/issues/60226) in rust and not specific to s390x. The best practice to fix this issue is to use `c_char` instead of `i8/u8` like [this](https://github.com/phanikishoreg/llvm-rs/commit/013d1e95406264df57879cda698dbe7a806b62b7).


jialez0 (approved):
LGTM.

sameo (approved):

PR #82, opened 2023-03-29
TDX | Workaround for old TDEL format used by TDVF now

jialez0:
Related issue: https://github.com/confidential-containers/attestation-service/issues/81


Xynnn007 (approved):
LGTM

PR #85, opened 2023-03-29
TDX verifier: revert ccel change

Xynnn007:
This change was applied because the given tdvf does not support CCEL. Now newest tdvf is brought and
CCEL is supported, thus we do not need this workground.


Xynnn007:
cc @jialez0

jialez0 (approved):

PR #86, opened 2023-03-31
TDX | Update CCEL parse logic to support both TDVF and TD-Shim

jialez0:
This change updates the parsing method of TDX CCEL, The TD SHIM specific cc-measurement library is no longer used, Use the eventlog-rs library instead to support CCEL generated by both TDVF and TD SHIM.

cc @Xynnn007

### DIFF
--- a/src/verifier/tdx/eventlog.rs
+++ b/src/verifier/tdx/eventlog.rs
@@ -1,28 +1,20 @@
 use anyhow::*;
 use byteorder::{LittleEndian, ReadBytesExt};
-use cc_measurement::{log::CcEvents, TPML_ALG_SHA384};
-use core::fmt;
 use core::mem::size_of;
-use sha2::{Digest, Sha384};
+use eventlog_rs::Eventlog;
 use std::convert::{TryFrom, TryInto};
 use std::string::ToString;
 
-/// Consist with https://github.com/confidential-containers/td-shim/blob/main/td-shim/src/event_log.rs
 #[derive(Debug, Clone, EnumString, Display)]
 pub enum MeasuredEntity {
     #[strum(serialize = "td_hob\0")]
-    TdHob,
+    TdShim,
     #[strum(serialize = "td_payload\0")]
-    TdPayload,
+    TdShimKernel,
     #[strum(serialize = "td_payload_info\0")]
-    TdPayloadInfo,
-    // From here down is not supported
-    #[strum(serialize = "td_payload_svn\0")]
-    TdPayloadSVN,
-    #[strum(serialize = "secure_policy_db")]
-    SecurePolicyDB,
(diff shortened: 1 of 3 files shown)


src/verifier/tdx/eventlog.rs
```diff
@@ -144,12 +89,12 @@ impl<'a> CcEventLog<'a> {
...
-        for (_, data) in self.cc_events {
-            if data.len() < event_desc_prefix.len() {
+        for event_entry in self.cc_events.log.clone() {
+            if event_entry.event_desc.len() < event_desc_prefix.len() {
                 continue;
             }
-            if &data[..event_desc_prefix.len()] == event_desc_prefix.as_slice() {
-                return Some(data.to_vec());
+            if &event_entry.event_desc[..event_desc_prefix.len()] == event_desc_prefix.as_slice() {
+                return Some(event_entry.event_desc);
```
Xynnn007, line 97: We can directly use `Some(..)` here without `return`

src/verifier/tdx/eventlog.rs
```diff
@@ -125,16 +73,13 @@ impl<'a> CcEventLog<'a> {
...
             }
-            if &data[..event_desc_prefix.len()] == event_desc_prefix.as_slice() {
-                let digest = &header.digest.digests[0];
-                if digest.hash_alg == TPML_ALG_SHA384 {
-                    let sha384 = digest.digest.sha384;
-                    return Some(hex::encode(sha384));
-                }
+            if &event_entry.event_desc[..event_desc_prefix.len()] == event_desc_prefix.as_slice() {
+                let digest = &event_entry.digests[0].digest;
+                return Some(hex::encode(digest));
```
Xynnn007, line 82: We can directly use `Some(..)` here without `return`

src/verifier/tdx/mod.rs
```diff
@@ -138,22 +134,20 @@ fn generate_parsed_claim(
...
                         "tdx-kernel-size{:?}",
                         uefi_platform_firmware_blob2.blob_length
                     );
-                    claim_map.insert(digest_name, serde_json::Value::String(td_paylod_digest));
-                }
-                _ => {
-                    warn!("parse CC EventLog: There is no tdx kernel measurement entry")
+                    claim_map.insert(digest_name, serde_json::Value::String(kernal_digest));
                 }
+                _ => todo!(),
```
Xynnn007, line 150: there is a `panic!` behind `todo!`, maybe we should use a `warn!` or something else here?

Xynnn007 (approved):
LGTM, some nits.. plz

src/verifier/tdx/eventlog.rs
jialez0: So as above.

PR #92, opened 2023-05-06
AS: Add API to set policy

jialez0:


src/policy_engine/mod.rs
```diff
@@ -27,4 +27,6 @@ pub trait PolicyEngine {
         reference_data_map: HashMap<String, Vec<String>>,
         input: String,
     ) -> Result<(bool, String)>;
+
+    fn set_policy(&mut self, policy_base64_string: String) -> Result<()>;
```
Xynnn007, line 31: I suggest change this API to `async`. And use `tokio::fs::write` to impl this to avoid blocking calling. How do you like?

src/policy_engine/opa/mod.rs
```diff
@@ -85,6 +85,13 @@ impl PolicyEngine for OPA {
 
         Ok((res_kv["allow"].as_bool().unwrap_or(false), res))
     }
+
+    fn set_policy(&mut self, policy: String) -> Result<()> {
+        let policy_bytes = base64::decode_config(policy, base64::URL_SAFE_NO_PAD)
+            .map_err(|_| anyhow!("Base64 decode OPA policy string failed"))?;
```
Xynnn007, line 91: Only comment: I like the detailed error information here : )

src/policy_engine/opa/mod.rs
```diff
@@ -85,6 +85,13 @@ impl PolicyEngine for OPA {
...
+        fs::write(&self.policy_file_path, policy_bytes)
+            .map_err(|e| anyhow!("Write OPA policy to file failed: {:?}", e))
+    }
 }
 
 #[cfg(test)]
```
Xynnn007, line 97: Could we have a test here if you like?

Xynnn007 (commented):
Thanks for the API! It is very useful!

niteeshkd:
It seems the API updates the existing opa policy file (e.g. .../opa/policy.rego). 
1) Can it be used to create a new policy file and use that in the policy evaluation? I think coco owner may provide different policy files for different cocos while using the same attestation service. 
2) If later some other type of policy engine is supported, is it expected that new API will be created to update the policy for that type of policy engine?

jialez0:
> It seems the API updates the existing opa policy file (e.g. .../opa/policy.rego).
> 
> 1. Can it be used to create a new policy file and use that in the policy evaluation? I think coco owner may provide different policy files for different cocos while using the same attestation service.
> 2. If later some other type of policy engine is supported, is it expected that new API will be created to update the policy for that type of policy engine?

@niteeshkd Thanks for proposal! I will try to think and design to implement the functions you mentioned.

jialez0:
@niteeshkd I have update the code, now set policy input specified the policy type and policy ID, you can take a look at this PR again. Thanks!
In this PR, we only use "default" as the policy ID for now, this is because how KBS/AS identifies coco resources is still under discussion and design. After we have a comprehensive mechanism to identify the coco resources, we can use different policy IDs based on that ID.

src/policy_engine/opa/mod.rs
```diff
@@ -23,36 +25,46 @@ pub struct GoString {
...
-            fs::create_dir_all(&policy_file_path)
+        policy_dir_path.push("opa");
+        if !policy_dir_path.as_path().exists() {
+            fs::create_dir_all(&policy_dir_path)
                 .map_err(|e| anyhow!("Create policy dir failed: {:?}", e))?;
         }
 
-        policy_file_path.push("policy.rego");
-        if !policy_file_path.as_path().exists() {
+        let mut default_policy_path = PathBuf::from(&policy_dir_path.to_str().unwrap());
```
Xynnn007, line 41: Please change the `unwrap` to an error handling here

src/policy_engine/opa/mod.rs
```diff
@@ -23,36 +25,46 @@ pub struct GoString {
...
+        default_policy_path.push("default.rego");
+        if !default_policy_path.as_path().exists() {
             let policy = std::include_str!("default_policy.rego").to_string();
-            fs::write(&policy_file_path, policy)?;
+            fs::write(&default_policy_path, policy)?;
```
Xynnn007, line 45: Would you like to add a new commit to change the `fs::write` to async (tokio?)

src/policy_engine/opa/mod.rs
```diff
@@ -23,36 +25,46 @@ pub struct GoString {
...
     ) -> Result<(bool, String)> {
-        let policy = fs::read_to_string(&self.policy_file_path)
+        let policy_file_path = format!(
+            "{}/{}.rego",
+            self.policy_dir_path
+                .to_str()
+                .ok_or_else(|| anyhow!("Miss Policy DirPath"))?,
+            policy_id.unwrap_or("default".to_string())
+        );
+        let policy = fs::read_to_string(policy_file_path)
```
Xynnn007, line 67: Would you like to add a new commit to change the fs::write to async (tokio?)

src/policy_engine/opa/mod.rs
```diff
@@ -85,6 +97,21 @@ impl PolicyEngine for OPA {
 
         Ok((res_kv["allow"].as_bool().unwrap_or(false), res))
     }
+
+    async fn set_policy(&mut self, input: SetPolicyInput) -> Result<()> {
+        if input.r#type != "opa" {
```
Xynnn007, line 102: Why do we need the `type` in a `SetPolicyInput` here? Does it mean a running attestation-service instance has multiple PolicyEngines escorting?

src/policy_engine/opa/mod.rs
```diff
@@ -85,6 +97,21 @@ impl PolicyEngine for OPA {
...
+            bail!("Policy Engine Type Unmatched");
+        }
+
+        let policy_bytes = base64::decode_config(input.policy, base64::URL_SAFE_NO_PAD)
+            .map_err(|e| anyhow!("Base64 decode OPA policy string failed: {:?}", e))?;
+        let mut policy_file_path = PathBuf::from(&self.policy_dir_path.to_str().unwrap());
+        policy_file_path.push(format!("{}.rego", input.policy_id));
+
+        tokio::fs::write(&policy_file_path, policy_bytes)
```
Xynnn007, line 111: We can change all `std::fs` into `tokio::fs` in this file.

src/policy_engine/opa/mod.rs
```diff
@@ -108,21 +135,42 @@ mod tests {
...
         assert!(res.unwrap().0 == false, "allow should be false");
     }
+
+    #[tokio::test]
+    async fn test_set_policy() {
+        let mut opa = OPA::new(PathBuf::from("./test_data")).unwrap();
+        let input = SetPolicyInput {
+            r#type: "opa".to_string(),
+            policy_id: "test".to_string(),
+            policy: "cGFja2FnZSBwb2xpY3kKCmltcG9ydCBmdXR1cmUua2V5d29yZHMuZXZlcnkKCmRlZmF1bHQgYWxsb3cgPSBmYWxzZQo=".to_string()
```
Xynnn007, line 171: Could you show the plaintext of the policy here and encode it into base64? This might help to understand the policy

src/policy_engine/opa/mod.rs
jialez0: I don't think we should change the `new()` method of OPA to aysnc...

src/policy_engine/opa/mod.rs
jialez0: This is to avoid setting the policy type of other engines to OPA, as the source of this policy is the remote caller. Therefore, I believe it is necessary to add a type field here to prevent such errors from occurring.

src/policy_engine/opa/mod.rs
jialez0: Why?

src/policy_engine/opa/mod.rs
```diff
@@ -85,6 +102,21 @@ impl PolicyEngine for OPA {
...
     }
+
+    async fn set_policy(&mut self, input: SetPolicyInput) -> Result<()> {
+        if input.r#type != "opa" {
+            bail!("Policy Engine Type Unmatched");
+        }
+
+        let policy_bytes = base64::decode_config(input.policy, base64::URL_SAFE_NO_PAD)
+            .map_err(|e| anyhow!("Base64 decode OPA policy string failed: {:?}", e))?;
+        let mut policy_file_path = PathBuf::from(&self.policy_dir_path.to_str().unwrap());
```
Xynnn007, line 113: could we replace the `unwrap` here with an error handling?

Xynnn007 (approved):

PR #95, opened 2023-05-22
github: Enable dependabot

sameo:
For automatic tracking of dependencies and version updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,23 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: daily
+    open-pull-requests-limit: 1
+    allow:
+      - dependency-type: direct
+      - dependency-type: indirect
+
+  - package-ecosystem: "gomod"
+    directory: "/attestation-service/src/cgo" # Location of shim's go.mod
+    schedule:
+      interval: "daily"
+    open-pull-requests-limit: 1
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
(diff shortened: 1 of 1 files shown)


Xynnn007 (approved):
lgtm, thanks!

Xynnn007 (commented):
We also need to track go libs because cgo is used
```yaml
# Dependencies listed in go.mod
  - package-ecosystem: "gomod"
    directory: "/attestation-service/src/cgo" # Location of shim's go.mod
    schedule:
      interval: "daily"
    open-pull-requests-limit: 1
```

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+- package-ecosystem: cargo
```
Xynnn007, line 3: Do we need an indent here? As https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#example-of-a-basic-setup-for-three-package-managers shows, the whole `package-ecosystem` object should be one of the list of values ​​of `updates`.

sameo: Fixed now. And I also added the go dependency, thanks for catching this @Xynnn007

dubek: I see in that link that there's also tracking for github actions:

```yaml
  # Maintain dependencies for GitHub Actions
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "weekly"
```

should we add that as well?

sameo: Good suggestion, done now. Thanks.

PR #105, opened 2023-06-13
TDX Verifier: Add fine grained parsed claims inside evidence

Xynnn007:
cc @jialez0 
Please take a look if it is suitable

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -141,6 +141,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "assert-json-diff"
+version = "2.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "47e4f2b81832e72834d7518d8487a0396a28cc408186a2e8854c0f98011faf12"
+dependencies = [
+ "serde",
+ "serde_json",
+]
+
+[[package]]
 name = "async-stream"
 version = "0.3.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


Xynnn007:
Note that there are two kinds of kernel parameters due to https://www.kernel.org/doc/html/v4.14/admin-guide/kernel-parameters.html:
- `A=B` style, like `console=hvc0`. Serialized to a key-value pair `"A":"B"` in JSON.
- `A` style, like `hw`. Serialized to a key-value pair with `null` value in JSON, s.t. `"A": null`

I handle the both in a `Map`, the format will look lile
```json
"kernel_parameters": {
    "console": "hvc0",
    "hw": null
}
```

If a parameter is `A=`, then we will serialize that into `"A": ""`, here the value is an empty string rather than `null`

sameo (approved):

jialez0:
@sameo Can we reverse this PR for now? Because if we merged this now, the OPA default policy can not parse the new format claims, this will cause the validation of the policy engine failure. Could we wait for this PR to add modifications to the default policy before merging it? So sorry for didn't remind that...

jialez0:
@sameo BTW, I think we should merge PR #106  with higher priority as it solves the problem of current AS being unable to verify evidence of real TDX hardware platform.

jialez0:
@Xynnn007 WDYT?

Xynnn007:
Yes. Related code is not prepared in the policy engine side. A better way is to revert this PR and wait for all other things are ok. Let me write an issue to track this.

Xynnn007:
Let me make another PR to provide a workaround. Better ways can be discussed in future

PR #106, opened 2023-06-15
Update TDX verifier dependencies to DCAP_1.16

jialez0:


attestation-service/Cargo.toml
```diff
@@ -28,7 +28,7 @@ chrono = { version = "0.4.19", features = [ "serde" ] }
 eventlog-rs = { version = "0.1.3", optional = true }
 futures = "0.3.17"
 hex = "0.4.3"
-intel-tee-quote-verification-rs = { version = "0.2.1", optional = true }
+sgx-dcap-quoteverify-rs = { git = "https://github.com/intel/SGXDataCenterAttestationPrimitives", tag = "DCAP_1.16", optional = true }
```
sameo, line 31: Could you please give an explanation on why we're moving from a TDX crate to an SGX/DCAP one? What does it bring? Is it backward compatible?

jialez0: This is actually a version rollback operation. It changed the name of `sgx-dcap-quoteverify` to `intel-tee-quoteverify` and published it on crates.io after Intel DCAP released version 1.16 for a period of time: https://github.com/intel/SGXDataCenterAttestationPrimitives/commit/2eb766ad819908cedcf5a672670cf9e176f2282a. However, if we directly use this new version of this crate package here, there will be errors when running, because the latest release versions of the related underlying dependencies (those we installed as .so files) are all 1.16, And the `intel-tee-quoteverify` is actually higher than version 1.16. So currently, in order for our code to run properly, we need to still use the dependency of version 1.16. After Intel releases version 1.17 of DCAP, I will update the version here.

sameo: Thanks a lot for the explanation @jialez0 . Could we please have a summary of that in the git commit message?

jialez0: > 

Sure, I'll add it.

jialez0:
@sameo Could you please take a look at this PR again?

sameo (approved):

PR #108, opened 2023-06-15
Update base64 to 0.21

sameo:
And fix a resolver issue.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,8 @@ members = [
     "bin/rvps-client",
 ]
 
+resolver = "2"
+
 [workspace.package]
 version = "0.1.0"
 authors = ["The Confidential Container Authors"]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -193,7 +193,7 @@ dependencies = [
  "assert-json-diff",
  "async-trait",
  "az-snp-vtpm",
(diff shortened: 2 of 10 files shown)


Cargo.toml
```diff
@@ -6,6 +6,7 @@ members = [
     "bin/grpc-as",
     "bin/rvps-client",
 ]
+resolver = "2"
```
mkulke, line 9: i'm curious, why is it necessary? I read it's implied by setting `edition = 2021`, which might be less obscure. but fwiw I don't get any compilation errors when removing that line.

fitzthum (approved):
Thanks. I was too lazy to do this when I added the snp verifier.

Xynnn007 (approved):
Thanks for this!

jialez0 (approved):

Cargo.toml
sameo: I see this linter warning with 1.72 when running a basic `cargo build`. Some of our dependencies are explicitly setting `edition = 2021` which implies `resolver = 2` which is different from the the virtual workspace (we do use a virtual workspace) default resolver.

I feel like I'm rephrasing the warning itself, not sure this is giving a better explanation...

mkulke (approved):

PR #110, opened 2023-06-20
TDX verifier: Refined the parsing content of tdx evidence

jialez0:


Xynnn007 (approved):
LGTM.

To simplify, this PR will flatten the parsed claims of different verifier drivers into a key-value map whose key is a "<TEE-TYPE>."-prefixed string. e.g. "tdx.quote.header.version"

For SNP, e.g. "snp.reported_tcb_bootloader"

This prefix helps to retrieve reference values in RVPS. Please take a look if this change looks good to you. cc @fitzthum @mkulke

attestation-service/src/utils.rs
```diff
@@ -23,12 +24,16 @@ use serde_json::{Map, Value};
...
+    tee: kbs_types::Tee,
+    claims: &TeeEvidenceParsedClaim,
+) -> Result<TeeEvidenceParsedClaim> {
     let mut map = Map::new();
+    let tee_type = to_variant_name(&tee)?;
     match claims {
         Value::Object(obj) => {
             for (k, v) in obj {
-                flatten_helper(&mut map, v, k.clone());
+                flatten_helper(&mut map, v, format!("{tee_type}.{}", k.clone()));
```
jepio, line 36: How about using `-` or `_` instead of `.`?

jialez0: In certain specific fields of SGX and TDX, they may already contain `-` or `_`. If we use them as delimiters, it may sometimes cause some confusion. Therefore, I think it is more appropriate for us to use `.` as the delimiter here. For the claims parsed internally by each TEE verifier, they can still use any format they like.

PR #114, opened 2023-06-22
Dockerfile: Copy TPM runtime dependencies

octaviansima:
TPM libraries installed during first build in Dockerfile are not available at runtime.

`apt-get install libtss2-dev` will install libraries in `/usr/lib/x86_64-linux-gnu` that the Rust executable needs to load. However, this step happens only during the stage in the build to create the Rust executable. We need to copy these dependencies to the final half of the build so the executable can find them.

Confirmed that the gRPC server now starts with a locally built image
```
[2023-06-22T18:51:20Z INFO  grpc_as::server] Listen socket: 0.0.0.0:50004
[2023-06-22T18:51:20Z INFO  grpc_as::server] Start a local RVPS (Server mode)
```

Fixes https://github.com/confidential-containers/attestation-service/issues/113

### DIFF
--- a/Dockerfile.as
+++ b/Dockerfile.as
@@ -14,14 +14,14 @@ RUN wget https://go.dev/dl/go1.20.1.linux-amd64.tar.gz && \
 ENV PATH="/usr/local/go/bin:${PATH}"
 
 # Install TPM Build Dependencies
-RUN apt-get update && apt install -y protobuf-compiler clang libtss2-dev
+RUN apt-get update && apt-get install -y protobuf-compiler clang libtss2-dev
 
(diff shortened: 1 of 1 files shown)


octaviansima:
@Xynnn007 can you help take a look at this? Noticed you were last to touch the Dockerfile

fitzthum:
Can you add a DCO sign-off to your commit message?

Xynnn007 (approved):
LGTM, thanks for the fix!

fitzthum (approved):
LGTM. Thanks and welcome to CoCo

PR #116, opened 2023-06-29
Update virtee/sev dep to 1.2

jepio:
Fix all the API breakage and update az-snp-vtpm dep for compat as well.

### DIFF
--- a/attestation-service/Cargo.toml
+++ b/attestation-service/Cargo.toml
@@ -39,7 +39,7 @@ scroll = { version = "0.11.0", default-features = false, features = ["derive"], 
 serde.workspace = true
 serde_json.workspace = true
 serde_variant = "0.1.2"
-sev = { version = "= 1.2.0", features = ["openssl", "snp"], optional = true }
+sev = { version = "1.2.0", features = ["openssl", "snp"], optional = true }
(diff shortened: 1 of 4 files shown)


jialez0 (approved):

jepio:
@jialez0 fixed cargo fmt and rebased due to conflicts in cargo.toml.

attestation-service/Cargo.toml
```diff
@@ -39,7 +39,7 @@ scroll = { version = "0.11.0", default-features = false, features = ["derive"],
 serde.workspace = true
 serde_json.workspace = true
 serde_variant = "0.1.2"
-sev = { version = "= 1.1.0", features = ["openssl"], optional = true }
+sev = { version = "= 1.2.0", features = ["openssl", "snp"], optional = true }
```
fitzthum, line 42: Do we still need to set an exact version with these changes?

fitzthum (approved):
LGTM.

attestation-service/Cargo.toml
jepio: no we don't, i've pushed this change

sameo (approved):

PR #121, opened 2023-07-28
Change the Attestation Results from a custom JSON structure to a JWT standard format.

jialez0:


README.md
```diff
@@ -87,22 +87,29 @@ The tee-evidence formats depend on the TEE and are typically defined by the each
...
+
+Claims format of the attestation results token is:
 
 ```json
 {
-    "tee": $tee_type,
-    "allow": true,
-    "output": {
-        "verifier_output": $verifier_output,
-        "policy_engine_output": $opa_output
```
Xynnn007, line 98: Do we need to mention that the original opa_output will be raised as an error information to the client side?

attestation-service/Cargo.toml
```diff
@@ -29,12 +29,15 @@ eventlog-rs = { version = "0.1.3", optional = true }
...
 sgx-dcap-quoteverify-rs = { git = "https://github.com/intel/SGXDataCenterAttestationPrimitives", tag = "DCAP_1.16", optional = true }
+jwt = { version = "0.16.0", features = ["openssl"]}
 kbs-types = "0.4"
 lazy_static = "1.4.0"
 log.workspace = true
 openssl = { version = "0.10.43", optional = true }
 path-clean = "1.0.1"
 prost.workspace = true
+rsa = { version = "0.9.2", features = ["sha2"] }
+rand = "0.8.5"
```
Xynnn007, line 40: Can you move the `rand` crate above `rsa` to follow dictionarial order?

attestation-service/src/config.rs
```diff
@@ -19,6 +20,13 @@ pub struct Config {
     pub policy_engine: String,
 
     pub rvps_store_type: StoreType,
+
+    /// The Attestation Result Token Broker type.
+    ///
+    /// Possible values:
+    /// * `Simple`
+    pub attestation_token_broker: AttestationTokenBrokerType,
+    pub attestation_token_config: AttestationTokenConfig,
```
Xynnn007, line 29: nit: add a empty line here

attestation-service/src/config.rs
```diff
@@ -1,3 +1,4 @@
+use crate::token::{AttestationTokenBrokerType, AttestationTokenConfig};
```
Xynnn007, line 1: nit: add a empty line here

attestation-service/src/config.rs
```diff
@@ -41,7 +51,11 @@ impl TryFrom<&Path> for Config {
     ///    {
     ///        "work_dir": "/var/lib/attestation-service/",
     ///        "policy_engine": "opa",
-    ///        "rvps_store_type": "LocalFs"
+    ///        "rvps_store_type": "LocalFs",
+    ///        "attestation_token_broker": "Simple",
+    ///        "attestation_token_config": {
+    ///            "duration_min": 5,
```
Xynnn007, line 57:
```suggestion
    ///            "duration_min": 5
```

attestation-service/src/lib.rs
```diff
@@ -17,19 +17,20 @@ extern crate strum_macros;
 pub mod config;
 pub mod policy_engine;
 pub mod rvps;
+mod token;
 mod utils;
 pub mod verifier;
 
+use crate::token::AttestationTokenBroker;
```
Xynnn007, line 24: nit: add a empty line here

attestation-service/src/lib.rs
```diff
@@ -134,11 +133,19 @@ impl AttestationService {
             .evaluate(reference_data_map, tcb.clone(), None)
             .await?;
 
-        let attestation_results =
-            AttestationResults::new(tee, result, None, Some(policy_engine_output), Some(tcb));
-        debug!("Attestation Results: {:?}", &attestation_results);
+        if !result {
+            return Err(anyhow!(
+                "Policy Engine verification failed: {policy_engine_output}"
+            ));
```
Xynnn007, line 139:
```suggestion
            bail!("Policy Engine verification failed: {policy_engine_output}");
```

attestation-service/src/token/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+
+const DEFAULT_TOKEN_TIMEOUT: i64 = 5;
+
+pub trait AttestationTokenBroker {
+    /// Issue an signed attestation token with custom claims.
+    /// Return base64 encoded Json Web Token.
+    fn issue(&self, custom_claims: Value) -> Result<String>;
+
+    /// Get the public keys and X.509 formatted certificate chain of the attestation token broker.
+    /// Returns the certificate chain in JWKS format (https://www.rfc-editor.org/rfc/rfc7517#appendix-B).
```
Xynnn007, line 21:
```suggestion
    /// Returns the certificate chain in [JWKS format](https://www.rfc-editor.org/rfc/rfc7517#appendix-B).
```

attestation-service/src/token/simple.rs
```diff
@@ -0,0 +1,105 @@
...
+        let exp = now + time::Duration::minutes(self.config.duration_min);
+
+        let mut claims = json!({
+            "iss": ISSUER_NAME,
+            "jwk": serde_json::from_str::<Value>(&self.pubkey_jwks()?)?["keys"][0].clone(),
+            "nbf": now.unix_timestamp(),
+            "exp": exp.unix_timestamp(),
+        })
+        .as_object()
+        .unwrap()
```
Xynnn007, line 66: `expect("...")` here is better I think. It can convey more information when the assertion here fails.

attestation-service/src/token/simple.rs
```diff
@@ -0,0 +1,105 @@
...
+        claims.extend(
+            custom_claims
+                .as_object()
+                .ok_or_else(|| anyhow!("Illegal token custom claims"))?
+                .to_owned(),
+        );
+
+        let claims_value = Value::Object(claims);
+        let claims_string =
+            URL_SAFE_NO_PAD.encode(&serde_json::to_string(&claims_value)?.as_bytes());
```
Xynnn007, line 78: Could you please bring `serde_json::to_string(&claims_value)?` outside to a single sentence? The nested callings is somehow hard to read

attestation-service/src/token/simple.rs
```diff
@@ -0,0 +1,105 @@
...
+
+        let signature_payload = format!("{header_string}.{claims_string}");
+        let signature = URL_SAFE_NO_PAD.encode(self.rs384_sign(signature_payload.as_bytes())?);
```
Xynnn007, line 81: ditto

bin/grpc-as/src/server.rs
```diff
@@ -108,13 +108,10 @@ impl AttestationService for Arc<RwLock<AttestationServer>> {
...
 
-        let results = serde_json::to_string(&attestation_results)
-            .map_err(|e| Status::aborted(format!("Parse attestation results: {e}")))?;
-
-        debug!("Attestation Results: {}", &results);
+        debug!("Attestation Token: {}", &attestation_token);
 
         let res = AttestationResponse {
-            attestation_results: results,
+            attestation_token: attestation_token,
```
Xynnn007, line 114:
```suggestion
            attestation_token,
```

Xynnn007 (commented):
Let me try to summrize this PR:

Move the token issurance from KBS to AS. Previously when the verification succeeds, an `AttestationResult` will be returned. After this PR, when the verification succeeds, a token will be returned. 

From my own point of view, the whole KBS-AS cluster now has three parts:
- Resource Keeper: Impled by KBS
- Token issuer: impled by KBS and move to AS by this PR
- Verifier: Impled by AS

A future goal in my mind is to decouple these three things. Maybe we can do this after merging `kbs` and `as` repo as they are both tenant-components and have dependency relationship.

attestation-service/src/token/simple.rs
jialez0: I will use `ok_or(|| ...)` instead :)

fitzthum (changes requested):
Code looks ok at first glance, but I am worried about this on a higher level. 

I don't think the AS can be solely responsible for generating the Attestation Token. This will not work in the case that the HW evidence is generic.

To be generally useful the attestation token should reflect the workload identity. Today the hw evidence is generic (by design). The identity of the workload cannot be derived from the hw evidence alone. Currently only the KBS knows what the identity of the workload is so the KBS should generate the certificate and include claims about the workload in it.

I know that there are some new proposals that would add details of the workload to the hw evidence, but these are not yet implemented and the plan is to support these alongside the existing method. I think this change would cause issues with the current trust model, so I think we should discuss it in much more depth before we do anything.

Does that make sense? Maybe I have misunderstood the PR.

jialez0:
> I don't think the AS can be solely responsible for generating the Attestation Token. This will not work in the case that the HW evidence is generic.
> 
> To be generally useful the attestation token should reflect the workload identity. Today the hw evidence is generic (by design). The identity of the workload cannot be derived from the hw evidence alone. Currently only the KBS knows what the identity of the workload is so the KBS should generate the certificate and include claims about the workload in it.

@fitzthum Why do you think only KBS knows what the identity of the workload? If we plan to inject the information of workload identity though some way like HOSTDATA mechanism from k8s control, the identity of the workload will be included in TEE evidence. But AS do not take sense of this field, AS's verification is generic, but this filed of workload information will be parsed from TEE evidence and included in AS's Attestation Token cliams. KBS will parse the claims of the Attestation Token and get the field of workload identity, then KBS uses policy engine to decide whether this workload have access to specific resources.

jialez0:
> I know that there are some new proposals that would add details of the workload to the hw evidence, but these are not yet implemented and the plan is to support these alongside the existing method.

@fitzthum Please give me the link of these proposal and I will see why those proposal make only KBS know the identity of workload.

jialez0:
> I don't think the AS can be solely responsible for generating the Attestation Token. This will not work in the case that the HW evidence is generic.

To be generally useful the attestation token should reflect the workload identity. Today the hw evidence is generic (by design). The identity of the workload cannot be derived from the hw evidence alone. Currently only the KBS knows what the identity of the workload is so the KBS should generate the certificate and include claims about the workload in it.

BTW, I don't think the Attestation Token should carrying identity authorization information, it is best not to confuse TEE general authorization with workload identity authorization. They should be controlled by different tokens. The token issued by AS only indicates that the software and hardware conditions of TEE are OK, and will not provide more information about identity. On the other word, the Attestation Token in this PR is equivalent to the JSON format we originally used, but with an additional signature protection and timeliness information.

fitzthum:
@jialez0 Ok, I think you're right that I am conflating some sort of workload identity certificate with the token from RCAR. I will review the code here later today.

attestation-service/src/token/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+    fn issue(&self, custom_claims: Value) -> Result<String>;
+
+    /// Get the public keys and X.509 formatted certificate chain of the attestation token broker.
+    /// Returns the certificate chain in [JWKS format](https://www.rfc-editor.org/rfc/rfc7517#appendix-B).
+    fn pubkey_jwks(&self) -> Result<String>;
+}
+
+#[derive(Deserialize, Debug, Clone, EnumString)]
+pub enum AttestationTokenBrokerType {
+    Simple,
```
fitzthum, line 27: What other types do you anticipate? Are we going to be proxying this trait to some other service at some point?

attestation-service/src/lib.rs
```diff
@@ -134,11 +134,17 @@ impl AttestationService {
             .evaluate(reference_data_map, tcb.clone(), None)
             .await?;
 
-        let attestation_results =
-            AttestationResults::new(tee, result, None, Some(policy_engine_output), Some(tcb));
```
fitzthum, line 138: It seems like we are no longer propagating `policy_engine_output` to the KBS. We do report it in the case of an error, but we might want to think about cases where the policy does not have a binary output. It is possible that a client might have different types of claims that map onto different types of resources available from the KBS. It seems like having the output of the policy reported to the KBS would be very useful in this type of situation.

fitzthum (commented):
Code looks good overall. A couple minor questions.

fitzthum (approved):

attestation-service/src/token/mod.rs
jialez0: Currently, we use a simple random RS384 key signature, but in real deployment scenarios, there may be more demands for the authority of AS.

For example, in a typical deployment scenario, KBS owner hands over AS to another owner he trust for deployment, and KBS owner needs to confirm that the verification result of TEE evidence must come from the AS he trust. In this case, there are stricter requirements for the content and signature of the token, and a new Token Broker trait object may need to be implemented to sign the token using the AS owner's private key, and attach detailed owner identity information and a trusted CA guaranteed public key certificate chain in the token.

The above is just an example. In practical productization, there may be flexibility. It is possible that the issuer of the token is an independent service center in the AS deployment environment, which is the case of remote connection you mentioned.

attestation-service/src/lib.rs
jialez0: Please refer to my [answer](https://github.com/confidential-containers/kbs/pull/139#issuecomment-1672505902) in your comment of https://github.com/confidential-containers/kbs/pull/139.

jialez0: @fitzthum Now I have modified the processing of the policy engine output to include the evaluation report in the Attestation Results Token :)

PR #131, opened 2023-08-28
snp: accept DER and non-DER encoding for hwID

katexochen:
There will be a fix happening upstream at the AMD KDS that switches the hwID value in the VCEK from a raw bytes value to a valid DER encoded octet string.


fitzthum (approved):
LGTM. Should we update the test VCEK at some point?

katexochen:
@fitzthum Yes, but I don't have one at hand with the new format.

larrydewey:
@katexochen @fitzthum I have added the fix in the [snpguest tool](https://github.com/virtee/snpguest/pull/22/files). I have included a copy of the new (corrected) format. We are not using the asn1 crate in `snpguest`, but the tests should be easily adjusted for the crate usage. Feel free to copy the data from it for your tests.

stefandeml:
Wait, how can this change upstream without updating the [VCEK specification](https://www.amd.com/en/developer/sev.html) and adding versioning on their endpoints?
There are companies using those services in production and rely on stable spec.
Can anyone share more details on this upstream change?

PR #146, opened 2023-10-08
Update Docker base image version and install openssl

jialez0:
Now we use ubuntu 20.04 as base image, the openssl version is 1.1.1 where the location of `libssl.so` is different from version 3.0+ and `csv-rs` crate now can only find `libssl.so` with version 3.0+


Xynnn007:
Will this change break the current AS/KBS?

Xynnn007:
test locally and it helps to build the image.

Xynnn007 (approved):

PR #148, opened 2023-10-16
Replace `from_utc` with `DateTime::from_naive_utc_and_offset`

chendave:
`DateTime::<Tz>::from_utc` is deprecated now.

see log here: https://github.com/confidential-containers/attestation-service/actions/runs/6531296026/job/17732254060?pr=147


chendave:
cc @Xynnn007

chendave:
seems like this should not be a independent change

PR #158, opened 2023-11-02
as: Update csv-rs dep to rev 9d8882e.

BaoshunFang:
Update csv-rs to be compatible with openssl-sys 0.9.94.

### DIFF
--- a/attestation-service/Cargo.toml
+++ b/attestation-service/Cargo.toml
@@ -30,7 +30,7 @@ cfg-if = "1.0.0"
 chrono = { version = "0.4.19", features = [ "serde" ] }
 codicon = { version = "3.0", optional = true }
 # TODO: change it to "0.1", once released.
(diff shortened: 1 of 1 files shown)


Xynnn007:
A updation of dockerfile is applied. Let's see whether it is ok. If not, we should merge in this release.

chendave:
Hit the same issue here: https://github.com/confidential-containers/attestation-service/actions/runs/6744102574/job/18333258311?pr=159

> error[E0412]: cannot find type `c_void` in this scope
  --> /root/.cargo/git/checkouts/csv-rs-6ce26198efa2f606/05fbacd/src/crypto/sm.rs:99:63
   |
99 |             EVP_DigestVerifyUpdate(mctx, msg.as_ptr() as *mut c_void, msg.len());
   |                                                               ^^^^^^ not found in this scope
   |


I think we should merge this? or else the code base is not compliable with `csv-verifier`.   @Xynnn007

PR #159, opened 2023-11-03
Cca: Get the evidence from EAR (EAT Attesation Result)[1]

chendave:
CCA validation by the Verasion has some overlapping with the RVPS, the similar validation has been done by the Verasion already. So the validation of evidence in CoCo is not needed in theory.

The parse of the CCA token here is to align with other verifier, e.g. TDX, and to support RVPS if that is the case of future planning.

NOTE that each of the value we get from EAR is base64 encoded hex value. So, the reference value that are registered for RVPS are expected to be base64 encoded hex value as well.

[1] https://datatracker.ietf.org/doc/draft-ietf-rats-ar4si/


chendave:
@Xynnn007 Is there any way for me to trigger the test again? or else can you help to rerun the test? thanks!

Xynnn007:
> @Xynnn007 Is there any way for me to trigger the test again? or else can you help to rerun the test? thanks!

I merged the csv fix. Just rebase the `upstream/main` and force push and the CI could be triggered again.

chendave:
rebased, let's see whether the CI is happy now.

